PR #4, opened 2019-06-28
Build on Windows.

jdm:
Completely untested, but the libraries now build at least.


openxr/Cargo.toml
```diff
@@ -26,6 +26,9 @@ libc = "0.2.50"
 shared_library = { version = "0.1.9", optional = true }
 void = "1"
 
+[target.'cfg(windows)'.dependencies]
+winapi = { version = "0.3", features = [] }
+
```
Ralith: Is this needed, given that we get the types from `sys::platform`?

PR #5, opened 2019-07-19
Add basic d3d11 bindings.

jdm:
This compiles but is untested so far.

### DIFF
--- a/openxr/src/graphics/d3d.rs
+++ b/openxr/src/graphics/d3d.rs
@@ -1,0 +1,103 @@
+use std::{mem, ptr};
+
+use sys::platform::*;
+
+use crate::*;
+
+/// The D3D11 graphics API
+///
+/// See [`XR_KHR_d3d11_enable`] for safety details.
+///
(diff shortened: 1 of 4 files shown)


openxr/src/graphics/d3d.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+}
+
+#[derive(Copy, Clone)]
+pub struct Requirements {
+    pub adapter_luid: LUID,
+    pub min_feature_level: D3D_FEATURE_LEVEL,
+}
+
+pub struct SessionCreateInfo {
```
Ralith: Any reason this shouldn't be `Copy + Clone`?

jdm: Not that I can see.

Ralith: Reasonable thing to expose by default, I think.

jdm: Fixed.

PR #12, opened 2019-11-08
Updated OpenXR SDK, added some QoL improvements

Limeth:
Updated OpenXR SDK, so that it can be built with Microsoft Visual Studio C++ Build Tools.
Fixed the build.rs build script.
Ignored target build directories of subcrates.
Added Cargo.lock to .gitignore.
Added a From tuple impl for Version.

Only tested on Windows.


Ralith:
Thanks!

PR #18, opened 2020-03-21
High level API for unknown extensions, update to 1.0.7

Ralith:


generator/src/main.rs
```diff
@@ -1350,17 +1350,19 @@ impl Parser {
...
             use crate::*;
 
             /// A subset of known extensions
-            #[derive(Debug, Copy, Clone, Eq, PartialEq, Default)]
+            #[derive(Debug, Clone, Eq, PartialEq, Default)]
             #[non_exhaustive]
             pub struct ExtensionSet {
                 #(#ext_set_fields)*
+                /// Extensions unknown to the high-level bindings
+                other: Vec<String>,
```
Manishearth, line 1365: Should this have been public?

Ralith: Whoops!

PR #24, opened 2020-04-23
Draft support for hand tracking

Manishearth:
I wrote some support for hand tracking using updated microsoft XML files. I wasn't quite sure how you'd want to structure this, features-wise, so I did it unconditionally as an example.

@Ralith would you like to take this PR and make it feature-conditional in the way you would like?


generator/src/main.rs
```diff
@@ -25,7 +25,7 @@ fn main() {
     args.next().unwrap();
     let mut source = File::open(args.next().map(PathBuf::from).unwrap_or_else(|| {
         Path::new(env!("CARGO_MANIFEST_DIR"))
-            .join("../sys/OpenXR-SDK/specification/registry/xr.xml")
+            .join("xr.xml")
```
Ralith, line 28: The commit that adds the new files should be squashed into this one to make it practical to review the changes in the generated output.

Ralith:
Sure, I can try hacking this up a bit. May not get to it for a few days though.

Ralith:
My initial thought is that we can probably modify the generator to feature-guard a blacklisted set of extensions, which will make a good foundation.

generator/src/main.rs
Manishearth: it is?

Manishearth:
Sounds good, I'm not in a particular rush. I managed to get this working and I mostly just wanted to have a proof of concept.

Blacklisting extensions could work, though we need to blacklist all of the associated types as well.

I was thinking we have two sets of generated files, one for the regular xr.xml and one for the Microsoft one (They'll have it in their repo as well soon, I think), and feature flags to flip between the two. `openxr` also has feature flags that toggle the hand tracking (and future other extension) types in tandem

Manishearth:
@Ralith any updates? if we can get the conditional compilation set up I can start working on the external APIs for most of the extension features :smile:

Ralith:
Been a bit busy, but it's still on my list!

Manishearth:
Good to hear!

In the meantime we're going to use a forked version, but hope to revert back once all the patches land.

Ralith:
Folded into #30. Please review!

PR #35, opened 2020-07-29
Add simple Vulkan example

Ralith:


openxr/examples/vulkan.rs
```diff
@@ -0,0 +1,649 @@
...
+            .next()
+            .expect("Vulkan device has no graphics queue");
+
+        let vk_device = vk_instance
+            .create_device(
+                vk_physical_device,
+                &vk::DeviceCreateInfo::builder()
+                    .queue_create_infos(&[vk::DeviceQueueCreateInfo::builder()
+                        .queue_family_index(queue_family_index)
+                        .queue_priorities(&[1.0])
```
msiglreith, line 151: I think this causes issues in release builds

edit: there are similar instances below as well

Ralith: Can you be more specific?

openxr/examples/vulkan.rs
```diff
@@ -0,0 +1,649 @@
...
+                    )
+                    .unwrap()
+            })
+            .collect::<Vec<_>>();
+
+        // Main loop
+        let mut swapchain = None;
+        let mut event_storage = xr::EventDataBuffer::new();
+        let mut session_running = false;
+        // Index of the current frame, wrapped by PIPELINE_DEPTH. Not to be confused with the
```
Stygeon, line 352: Love this comment.

openxr/examples/vulkan.rs
```diff
@@ -0,0 +1,649 @@
...
+                .vulkan_graphics_device(system, vk_instance.handle().as_raw() as _)
+                .unwrap() as _,
+        );
+
+        let queue_family_index = vk_instance
+            .get_physical_device_queue_family_properties(vk_physical_device)
+            .into_iter()
+            .enumerate()
+            .filter_map(|(queue_family_index, info)| {
+                if info.queue_flags.contains(vk::QueueFlags::GRAPHICS) {
```
Stygeon, line 136: I don't see another queue anywhere - don't you have to check for presentation capabilities? I assume this is not an issue here and related to the xr part.

openxr/examples/vulkan.rs
```diff
@@ -0,0 +1,649 @@
...
+                }
+            }
+
+            if !session_running {
+                // Don't grind up the CPU
+                std::thread::sleep(Duration::from_millis(100));
+                continue;
+            }
+
+            let state = frame_wait.wait().unwrap();
```
Stygeon, line 406: Maybe using an xr prefix could make this clearer. This confused me for a hot second before I realised it's not a vulkan helper. Since the vulkan stuff is mostly prefixed, maybe prefixing the xr stuff would be more consistent.

openxr/examples/vulkan.rs
```diff
@@ -0,0 +1,649 @@
...
+                    .end(
+                        state.predicted_display_time,
+                        xr::EnvironmentBlendMode::OPAQUE,
+                        &[],
+                    )
+                    .unwrap();
+                continue;
+            }
+
+            if swapchain.is_none() {
```
Stygeon, line 420: Very nice idea to use an option type for the swap chain. Makes it clear how to signal that it should be recreated.

openxr/examples/vulkan.rs
msiglreith: Sry! build() will discard the lifetime information and then it is not ensured that the temporary array [1.0] will live long enough, I believe. At least I faced this issue a few times already in prior rust/ash version. In release mode this tend to cause issues then as the memory gets reused.

openxr/examples/vulkan.rs
```diff
@@ -0,0 +1,649 @@
...
+                                    &vk::ImageViewCreateInfo::builder()
+                                        .image(color_image)
+                                        .view_type(vk::ImageViewType::TYPE_2D)
+                                        .format(COLOR_FORMAT)
+                                        .subresource_range(vk::ImageSubresourceRange {
+                                            aspect_mask: vk::ImageAspectFlags::COLOR,
+                                            base_mip_level: 0,
+                                            level_count: 1,
+                                            base_array_layer: 0,
+                                            layer_count: 2,
```
Stygeon, line 464: 2 for left and right viewports, I assume?

openxr/examples/vulkan.rs
```diff
@@ -0,0 +1,649 @@
...
+                        .collect(),
+                });
+            }
+            let swapchain = swapchain.as_mut().unwrap();
+            let image_index = swapchain.handle.acquire_image().unwrap();
+            swapchain.handle.wait_image(xr::Duration::INFINITE).unwrap();
+
+            // Ensure the last use of this frame's resources is 100% done
+            vk_device
+                .wait_for_fences(&[fences[frame]], true, !0)
```
Stygeon, line 491: Probably nitpicking since !0 should be clear enough, but maybe using an explicit max value would be even clearer.

openxr/examples/vulkan.rs
```diff
@@ -0,0 +1,649 @@
...
+            let scissors = [vk::Rect2D {
+                offset: vk::Offset2D { x: 0, y: 0 },
+                extent: swapchain.resolution,
+            }];
+            vk_device.cmd_set_viewport(cmd, 0, &viewports);
+            vk_device.cmd_set_scissor(cmd, 0, &scissors);
+
+            // Draw the scene. Multiview means we only need to do this once, and the GPU will
+            // automatically broadcast operations to all views. Shaders can use `gl_ViewIndex` to
+            // e.g. select the correct view matrix.
```
Stygeon, line 537: Nice comments.

openxr/examples/vulkan.rs
Ralith: Indeed, the OpenXR runtime is responsible for shuffling the swapchain image wherever it needs to be for the user to actually see it.

openxr/examples/vulkan.rs
Ralith: `build()` below is mandatory because we're populating an array of `DeviceQueueCreateInfo`s. The use of temporaries here and below is sound because temporaries [are guaranteed to outlive the statement they occur in](https://doc.rust-lang.org/reference/expressions.html#temporaries), which is the only context in which they are accessed.

openxr/examples/vulkan.rs
Ralith: Good thought.

openxr/examples/vulkan.rs
Ralith: Made this a little more explicit.

PR #38, opened 2020-08-08
Fix android

Jake-Shadle:
The `jobject` type is part of the signature of an Android function, this just adds a dependency on the `jni` crate and reexports that type so that Android can compile successfully.

Let me know if you want me to add a build check for Android in CI.


Ralith:
Great catch, thanks! CI coverage for Android would be very welcome; this is certainly one of the highest-risk areas for future silent breakage otherwise.

Jake-Shadle:
Cool, I'll do that tomorrow then.

Jake-Shadle:
Ok, I _think_ that should work to build android on ubuntu, but skip tests, but the actions being run for this PR are from the master branch so can't say for certain.

PR #48, opened 2020-09-28
Vulkan example tweaks

str4d:
Includes a couple of small tweaks to improve the robustness of the example.

I also ported over some documentation from @maluoi's C++ example code, that was helpful for me in working out what was going on in this implementation. I think this is acceptable usage (both examples are MIT-licensed), but @maluoi LMK if you'd prefer this not be included.

### DIFF
--- a/openxr/examples/vulkan.rs
+++ b/openxr/examples/vulkan.rs
@@ -76,7 +76,7 @@ fn main() {
     // Check what blend mode is valid for this device (opaque vs transparent displays). We'll just
     // take the first one available!
     let environment_blend_mode = xr_instance
-        .enumerate_environment_blend_modes(system, xr::ViewConfigurationType::PRIMARY_STEREO)
+        .enumerate_environment_blend_modes(system, VIEW_TYPE)
         .unwrap()[0];
 
     // OpenXR wants to ensure apps are using the correct graphics card, so the renderer MUST be set
@@ -412,9 +412,7 @@ fn main() {
                         println!("entered state {:?}", e.state());
                         match e.state() {
                             xr::SessionState::READY => {
-                                session
-                                    .begin(xr::ViewConfigurationType::PRIMARY_STEREO)
-                                    .unwrap();
+                                session.begin(VIEW_TYPE).unwrap();
                                 session_running = true;
                             }
                             xr::SessionState::STOPPING => {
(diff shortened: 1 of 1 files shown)


openxr/examples/vulkan.rs
```diff
@@ -409,27 +443,41 @@ fn main() {
...
                         &[],
                     )
                     .unwrap();
                 continue;
             }
 
             if swapchain.is_none() {
+                // Now we need to find all the viewpoints we need to take care of! For a stereo
+                // headset, this should be 2. Similarly, for an AR phone, we'll need 1, and a VR
+                // cave could have 6, or even 12!
```
Ralith, line 467: I feel like this is a little misleading, as it could be understood to mean that the number of views is unpredictable, whereas in fact it's a property of the view configuration type--so here we're absolutely guaranteed that there are exactly two.

openxr/examples/vulkan.rs
```diff
@@ -409,27 +443,41 @@ fn main() {
...
+                //
+                // Because we are using multiview in this example, we will only look at the first
+                // viewpoint, and assume that the other viewpoints are identical.
```
Ralith, line 470: Maybe we should have an explicit assert rather than just documenting the assumption?

Ralith (changes requested):
Thanks!

openxr/examples/vulkan.rs
```diff
@@ -440,14 +488,17 @@ fn main() {
                         usage_flags: xr::SwapchainUsageFlags::COLOR_ATTACHMENT
                             | xr::SwapchainUsageFlags::SAMPLED,
                         format: COLOR_FORMAT.as_raw() as _,
-                        sample_count: 1,
+                        sample_count: views[0].recommended_swapchain_sample_count,
```
Ralith, line 491: The Vulkan graphics pipeline we create is not set up for multisampling, so we should either leave this hardcoded to 1 or pursue a proper multisampling setup. I don't think there's any compatibility hazard for leaving it at 1.

str4d: Aha, that makes sense. I'll revert this and add documentation to this effect.

str4d:
Addressed comments from @Ralith.

openxr/examples/vulkan.rs
```diff
@@ -409,27 +441,40 @@ fn main() {
...
                 continue;
             }
 
             if swapchain.is_none() {
+                // Now we need to find all the viewpoints we need to take care of! This is a
+                // property of the view configuration type; in this example we use PRIMARY_STEREO,
+                // so we should have 2 viewpoints.
+                //
+                // Because we are using multiview in this example, we will only look at the first
+                // viewpoint, and assume that the other viewpoints are identical.
```
Ralith, line 468:
```suggestion
                // Because we are using multiview in this example, we require that all view
                // dimensions are identical.
```

Ralith (changes requested):

Ralith:
Thanks!

maluoi:
Super glad that code was helpful for you! And I love that even the comments can get a bit of re-use :)

PR #51, opened 2020-10-02
Update to OpenXR 1.0.12

Ralith:


Ralith:
@Jake-Shadle can you advise on whether we can come up with something safer than `*mut c_void` for the `LoaderInitInfoAndroidKHR` builder's parameters?

Jake-Shadle:
I don't know I'm not really familiar with the code in this repo or in the native openxr library, can it not be type aliased like the other c_void types? It could also be done as an opaque struct if it's just used as an opaque pointer from the native side.

Ralith:
Upstream OpenXR just refers to them as `void*`, which is what this PR exposes in the -sys API. I'd like to also have safe high-level bindings, but I don't have any context for Android to know if that's possible. I'll probably just blacklist them from the high-level bindings for now.

Jake-Shadle:
That seems reasonable, if someone actually needs it it can always be added later. :slightly_smiling_face:

PR #52, opened 2020-10-09
More Robust Null Terminator Handling

### LINKED ISSUE #47: get_str doesn't reliably strip null bytes
In #46 I noted the following crash in the `vulkan` example:
```
     Running `C:\Users\me\Documents\dev\rust\sat\openxrs\target\debug\examples\vulkan.exe`
loaded OpenXR runtime: Oculus 1.52.0
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: NulError(38, [86, 75, 95, 75, 72, 82, 95, 103, 101, 116, 95, 112, 104, 121, 115, 105, 99, 97, 108, 95, 100, 101, 118, 105, 99, 101, 95, 112, 114, 111, 112, 101, 114, 116, 105, 101, 115, 50, 0])', openxr\examples\vulkan.rs:65:34
```
It turns out that this is because for both `Instance::vulkan_instance_extensions` and `Instance::vulkan_device_extensions`, `get_arr` is returning an array with two trailing null bytes, but `get_str` assumes that the string will only contain one, and simply truncates a single byte.

I can temporarily fix this in the example with `trim_end_matches('\x00')`, but we should either understand where the extra null byte is coming from (and remove it), or alter `get_str` to strip as many null bytes as exist.

bigmstone:
In most recent SteamVR beta OpenXR calls can return values with multiple
null terminators. This will cause the existing solution to remove only
the final terminator. This commit adds a more robust check to truncate
on the first occurrence of null terminator.

Closes #47

### DIFF
--- a/openxr/src/lib.rs
+++ b/openxr/src/lib.rs
@@ -82,8 +82,10 @@ fn fixed_str_bytes(x: &[c_char]) -> &[u8] {
 
 fn get_str(mut getter: impl FnMut(u32, &mut u32, *mut c_char) -> sys::Result) -> Result<String> {
     let mut bytes = get_arr(|x, y, z| getter(x, y, z as _))?;
-    // Strip null byte
-    bytes.truncate(bytes.len() - 1);
+    // Truncate at first null byte
+    let first_nt = bytes.iter().rposition(|&x| x != 0).expect("missing null") + 1;
+    bytes.truncate(first_nt);
+
     unsafe { Ok(String::from_utf8_unchecked(bytes)) }
 }
 


openxr/src/lib.rs
```diff
@@ -82,8 +82,16 @@ fn fixed_str_bytes(x: &[c_char]) -> &[u8] {
...
     let mut bytes = get_arr(|x, y, z| getter(x, y, z as _))?;
-    // Strip null byte
-    bytes.truncate(bytes.len() - 1);
+    // Terminate at first null byte
+    let mut first_nt = bytes.len();
+    for (index, byte) in bytes.iter().enumerate() {
+        if *byte == 0 && index <= first_nt {
+            first_nt = index;
+        }
+    }
```
Ralith, line 91:
```suggestion
    let mut first_nt = bytes.iter().rfind(|&&x| x != 0).expect("missing null") + 1;
```

Ralith (changes requested):
Steam too, now? Argh. Thanks.

openxr/src/lib.rs
bigmstone: Since it's Double Ended Iterator this would return the actual byte instead of the index. Additionally, assuming index was returned,  this would break after the first occurrence from the rear leaving the last one in place. I've updated my implementation to use `position` which have confirmed to work, but is still satisfyingly terse.

bigmstone:
@Ralith this wasn't historically a problem but started in SteamVR beta `1.15.2`

openxr/src/lib.rs
Ralith: Whoops, right, should be `rposition(|&x| ...)`

openxr/src/lib.rs
```diff
@@ -82,8 +82,10 @@ fn fixed_str_bytes(x: &[c_char]) -> &[u8] {
 
 fn get_str(mut getter: impl FnMut(u32, &mut u32, *mut c_char) -> sys::Result) -> Result<String> {
     let mut bytes = get_arr(|x, y, z| getter(x, y, z as _))?;
-    // Strip null byte
-    bytes.truncate(bytes.len() - 1);
+    // Truncate at first null byte
+    let first_nt = bytes.iter().position(|&x| x == 0).expect("missing null");
```
Ralith, line 86:
```suggestion
    let first_nt = bytes.iter().rposition(|&x| x != 0).expect("missing null") + 1;
```
Makes more sense to search from the end.

Ralith (approved):
Thanks!

PR #53, opened 2020-10-11
Check that required Vulkan extensions exist in vulkan example

str4d:
The example still panics if an extension is missing, but now instead of
seeing ERROR_EXTENSION_NOT_PRESENT in the panic message, the user now
sees which extension is missing.

### DIFF
--- a/openxr/examples/vulkan.rs
+++ b/openxr/examples/vulkan.rs
@@ -6,7 +6,7 @@
 //! largely decouple its Vulkan and OpenXR components and handle errors gracefully.
 
 use std::{
-    ffi::CString,
+    ffi::{CStr, CString},
     io::Cursor,
     sync::{
         atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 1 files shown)


str4d:
I added this code to the example in order to debug the `ERROR_EXTENSION_NOT_PRESENT`, which revealed that it was caused by SteamVR specifying `VT_EXT_debug_marker` (which I could then search for, uncovering [threads like these](https://steamcommunity.com/app/250820/discussions/8/2950376208547832257/)). I figure just including this in the example saves others the first step.

str4d:
And for future devs who encounter this, the patch to enable SteamVR (beta 1.15.2) to run the example (in addition to #52):
```diff
diff --git a/openxr/examples/vulkan.rs b/openxr/examples/vulkan.rs
index 34ac623..4e40a72 100644
--- a/openxr/examples/vulkan.rs
+++ b/openxr/examples/vulkan.rs
@@ -147,6 +147,7 @@ fn main() {
         .vulkan_device_extensions(system)
         .unwrap()
         .split(' ')
+        .filter(|&ext| ext != "VK_EXT_debug_marker")
         .map(|x| CString::new(x).unwrap())
         .collect::<Vec<_>>();
     println!("required Vulkan device extensions: {:?}", vk_device_exts);
```

Ralith (approved):
Thanks! I'm actually open to including the workaround directly, though hopefully it'll only be needed temporarily...

PR #73, opened 2021-02-17
Add example demonstratring how to get tracking data

exrook:
Minimal example that gets pose information from tracked controllers.

It does not render anything but still makes a Vulkan instance based on the vulkan.rs example.
Theoretically we could use XR_MND_headless extension if it is present but my system does not support it, nor do the current bindings.

### DIFF
--- a/openxr/examples/vulkan.rs
+++ b/openxr/examples/vulkan.rs
@@ -85,20 +85,17 @@ fn main() {
     // extensions, so the instance and device MUST be set up before Instance::create_session.
 
     let vk_target_version = vk::make_version(1, 1, 0); // Vulkan 1.1 guarantees multiview support
+    let vk_target_version_xr = xr::Version::new(1, 1, 0);
 
     let reqs = xr_instance
         .graphics_requirements::<xr::Vulkan>(system)
         .unwrap();
-    if reqs.min_api_version_supported
-        > xr::Version::new(
-            vk::version_major(vk_target_version) as u16,
-            vk::version_minor(vk_target_version) as u16,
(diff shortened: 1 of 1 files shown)


Ralith:
Rather than duplicating a significant amount of code, could we just add this to the existing example?

exrook:
Sure, I can change it to that.
I just found the 600+ line vulkan example a little intimidating when I first found this project, and since I was only interested in getting the tracking data out I thought it would be nice to have a less involved example. Although at 400+ lines I suppose it's debatable how well this example fulfills that goal.

Ralith:
Yeah, if it's going to be super verbose anyway--and Vulkan always is--then we might as well consolidate and save maintenance effort. Alternatively, I'd also be happy with a more focused example that targets headless mode, though indeed it needs work at the wrapper layer. I'm equipped/heappy to test against monado, if you're interested in working on that.

exrook:
I've unified the examples now.

I may add a headless example in a future PR. I actually just tried installing monado on my laptop and it surprisingly (compared to SteamVR) just worked, although getting the vive working properly seems like it will require more effort, the example just hangs with the vive plugged in (and monado shows the vr view on my monitor instead of the hmd).

At the moment however I'm interested in getting an actual 3d example working, with proper view projection to the headset and showing the controller location/pose as cubes, so that may also come in a future PR. ...then after that I'll look at addressing https://github.com/gfx-rs/gfx/issues/3219 because "raw" vulkan really is obnoxious to work with.

Ralith (approved):
Thanks!

PR #74, opened 2021-02-21
Fix example Vulkan version check

Ralith:
From the spec:
> maxApiVersionSupported is the maximum Vulkan Instance API version that the runtime has been tested on and is known to support. Newer Vulkan Instance API versions might work if they are compatible.

cc @exrook


exrook:
lgtm

PR #75, opened 2021-02-22
Update vulkan example

exrook:
Example now renders in 3D using the tracked HMD position and shows cubes at the location of each hand controller.


openxr/examples/vulkan.rs
```diff
@@ -382,6 +438,151 @@ fn main() {
             .create_reference_space(xr::ReferenceSpaceType::STAGE, xr::Posef::IDENTITY)
             .unwrap();
 
+        let vk_memory_properties =
+            unsafe { vk_instance.get_physical_device_memory_properties(vk_physical_device) };
+
+        // Setup vertex buffer
+        let verticies: Vec<_> = Vertex::cube([1.0, 0.0, 0.0]).collect();
```
Ralith, line 445:
```suggestion
        let vertices: Vec<_> = Vertex::cube([1.0, 0.0, 0.0]).collect();
```

openxr/examples/vulkan.rs
```diff
@@ -382,6 +438,151 @@ fn main() {
...
+                    vk::MemoryMapFlags::empty(),
+                )
+                .unwrap();
+            let mut align = ash::util::Align::new(
+                data_ptr,
+                mem::align_of::<f32>() as u64,
+                vertex_buffer_mem_requirements.size,
+            );
+            align.copy_from_slice(&verticies);
+            vk_device.unmap_memory(vertex_memory);
```
Ralith, line 583:
```suggestion
```
There's rarely any need to unmap memory.

openxr/examples/vulkan.rs
```diff
@@ -382,6 +438,151 @@ fn main() {
...
+                    0,
+                    vertex_buffer_size as u64,
+                    vk::MemoryMapFlags::empty(),
+                )
+                .unwrap();
+            let mut align = ash::util::Align::new(
+                data_ptr,
+                mem::align_of::<f32>() as u64,
+                vertex_buffer_mem_requirements.size,
+            );
```
Ralith, line 581: `Align` isn't doing anything useful here.

openxr/examples/vulkan.rs
```diff
@@ -670,6 +961,29 @@ fn main() {
...
+            // Copy the view transforms to the GPU
+            {
+                let data_ptr = vk_device
+                    .map_memory(
+                        uniform_buffers[frame].1,
+                        0,
+                        uniform_buffer_size as u64,
+                        vk::MemoryMapFlags::empty(),
+                    )
+                    .unwrap();
```
Ralith, line 975: We shouldn't map, let alone unmap, every frame.

openxr/examples/vulkan.rs
```diff
@@ -772,5 +1091,101 @@ struct Framebuffer {
...
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug)]
+#[repr(C)]
+struct CameraTransform<S> {
+    view: Matrix4<S>,
+    project: Matrix4<S>,
+}
```
Ralith, line 1164: This doesn't need to be generic.

openxr/examples/vulkan.rs
```diff
@@ -772,5 +1091,101 @@ struct Framebuffer {
...
+
+impl Vertex {
+    pub fn cube(color: [f32; 3]) -> impl Iterator<Item = Self> {
+        [1, 0, 3, 2, 6, 0, 4, 1, 5, 3, 7, 6, 5, 4]
+            .into_iter()
+            .map(move |i| Self {
+                pos: [(i & 1) as f32, ((i >> 1) & 1) as f32, ((i >> 2) & 1) as f32],
+                color: color.clone(),
+            })
+    }
```
Ralith, line 1109: It might be a lot more concise to hardcode the geometry into the vertex shader (indexed using `gl_VertexIndex`) rather than using a conventional vertex buffer that we hardcode data into anyway.

openxr/examples/fullscreen.vert
```diff
@@ -1,8 +1,27 @@
...
 layout(location = 0) out vec2 screen_coords;
+layout(location = 1) out vec3 vColor;
+
+layout(push_constant) uniform ModelTransform {
+    mat4 model;
+};
+
+struct Transform {
+    mat4 view;
+    mat4 proj;
```
Ralith, line 15: If we don't need these to be separate, they should be multiplied together in advance on the CPU.

openxr/examples/fullscreen.vert
```diff
@@ -1,8 +1,27 @@
...
+    Transform transform[2];
+};
 
 void main()  {
-    screen_coords = vec2((gl_VertexIndex << 1) & 2, gl_VertexIndex & 2);
-    gl_Position = vec4(screen_coords * 2.0f + -1.0f, 0.0, 1.0f);
+    vec4 world_pos = model * vec4(vPosition, 1.0);
+    vec4 pos = transform[gl_ViewIndex].proj * transform[gl_ViewIndex].view * world_pos;
+    gl_Position = pos;
+    vColor = vec3(gl_VertexIndex/10.0, world_pos.gb);
```
Ralith, line 26: If this is no longer emitting a fullscreen triangle, the file should be renamed.

Ralith (changes requested):

openxr/examples/fullscreen.vert
exrook: Renamed to transform.vert

openxr/examples/vulkan.rs
exrook: What do you recommend instead? I did it this way because this is what I saw in the ash examples. (and I changed it again to use `vertex_buffer_mem_requirements.alignment` for the second argument in the force push I just did but I suspect that may have been incorrect after doing some more research)

Ralith: You shouldn't need to do anything special at all. Memory allocated according to the requirements of a vertex buffer will necessarily satisfy the buffer's GPU-side alignment requirements. I believe the CPU-side alignment should also be implicitly satisfied. Writing data starting at any location other than the start of the buffer, or with a non-dense stride, would constitute a mismatch with the vertex attribute layout declared in the pipeline.

openxr/examples/vulkan.rs
exrook: I agree it would be more concise to hardcode the vertices (and how I'm generating these was inspired by what the vertex shader was previously doing), but I think for the example it may make sense to show something closer to the average use case where a user will want to upload their models into a vertex buffer.

openxr/examples/vulkan.rs
Ralith: (unfortunately the ash examples, and indeed most introductory vulkan examples, aren't great; there's a PR in need of review to improve them)

exrook: Would it be sound to construct an `&mut [T]` from gpu memory or should I always go through ptr:;write?

exrook: I see `Align` does create a slice but my concern is that the data in the buffer may not be valid for reads. I think I'll use `std::ptr::copy` for now.

Ralith: Creating a slice to uninitialized memory would be unsound, yeah. [`copy_nonoverlapping`](https://doc.rust-lang.org/std/ptr/fn.copy_nonoverlapping.html) should be preferred.

exrook: Sounds good

PR #79, opened 2021-05-15
Support custom extensions and next struct

blaind:
This pull request is part of having OpenXR rendering for Bevy through WGPU-rs at https://github.com/blaind/xrbevy

For current usage, see code at https://github.com/blaind/bevy/blob/0ba8d6e4c3aaf37573ab0cbf774d8dbf63992374/crates/bevy_openxr/src/platform/oculus_android/mod.rs#L48

* InstanceCreateInfoAndroidKHR is supplied for next argument
* Tried the other_extensions part (for enabling custom Oculus Quest 2 extensions), but did not manage to implement it yet

### DIFF
--- a/openxr/src/entry.rs
+++ b/openxr/src/entry.rs
@@ -1,7 +1,8 @@
 #[cfg(feature = "loaded")]
 use libloading::Library;
 use std::{
-    ffi::{CStr, CString},
+    borrow::Cow,
+    ffi::{c_void, CStr, CString},
     mem, ptr,
     sync::Arc,
 };
@@ -116,6 +117,8 @@ impl Entry {
         &self,
         app_info: &ApplicationInfo,
         required_extensions: &ExtensionSet,
+        other_extensions: Option<Vec<Cow<[u8]>>>,
+        next: Option<*const c_void>,
         layers: &[&str],
     ) -> Result<Instance> {
         assert!(
@@ -128,7 +131,12 @@ impl Entry {
             "engine names are limited to {} bytes",
             sys::MAX_ENGINE_NAME_SIZE
         );
-        let ext_names = required_extensions.names();
+        let mut ext_names = required_extensions.names();
+
+        if let Some(other_extensions) = other_extensions {
+            ext_names.extend(other_extensions);
+        }
+
         let ext_ptrs = ext_names
             .iter()
             .map(|x| x.as_ptr() as *const _)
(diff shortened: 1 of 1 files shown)


Ralith:
Hi, and thanks for the PR!

> InstanceCreateInfoAndroidKHR is supplied for next argument

This makes the API unsafe. At the `openxr` level, I'd prefer to provide explicit, as-safe-as-possible support for specific desired extensions. If you need extensions that don't have a wrapper yet and don't want to roll one yourself, you can use [`Entry::fp()`](https://docs.rs/openxr/0.14.0/openxr/struct.Entry.html#method.fp) to get the raw FFI entry point and call that yourself with appropriate data, similar to your example usage. Would that address your requirements?

> Tried the other_extensions part (for enabling custom Oculus Quest 2 extensions), but did not manage to implement it yet

This is already exposed as the `other` member of [`ExtensionSet`](https://docs.rs/openxr/0.14.0/openxr/struct.ExtensionSet.html).

blaind:
> Hi, and thanks for the PR!

Thank you for a quick reply! :)

> > InstanceCreateInfoAndroidKHR is supplied for next argument
> 
> This makes the API unsafe. At the `openxr` level, I'd prefer to provide explicit, as-safe-as-possible support for specific desired extensions. If you need extensions that don't have a wrapper yet and don't want to roll one yourself, you can use [`Entry::fp()`](https://docs.rs/openxr/0.14.0/openxr/struct.Entry.html#method.fp) to get the raw FFI entry point and call that yourself with appropriate data, similar to your example usage. Would that address your requirements?

Sounds reasonable. In future, are you planning to support platform-specific initialization code, or should that be left for implementations?

> > Tried the other_extensions part (for enabling custom Oculus Quest 2 extensions), but did not manage to implement it yet
> 
> This is already exposed as the `other` member of [`ExtensionSet`](https://docs.rs/openxr/0.14.0/openxr/struct.ExtensionSet.html).

Its been a while since I tested this, but if I remember correctly, I had a problem where the `extensionset.next`, relating to the pointers not being persistent. But it seems you're wrapping them into `Cow<'static', ..>` so most probably I remember wrongly.

Ralith:
> In future, are you planning to support platform-specific initialization code, or should that be left for implementations?

I'm happy to review PRs to add high-level interfaces for OpenXR extensions, including platform specific ones. There's already a few cases of this, such as the OpenGL session create info. `cfg` guards may be called for.

> I had a problem where the extensionset.next, relating to the pointers not being persistent.

`ExtensionSet` has no field or method named `next`, so you're probably thinking of something else.

PR #83, opened 2021-06-27
Add android support

zarik5:
This is inspired from @blaind's work: https://github.com/blaind/bevy_openxr/blob/main/crates/bevy_openxr/src/platform/oculus_android/mod.rs

While @blaind work is known to be working, I didn't provide any testing for this code. If you want, I'm ok for waiting to have a working example before merging.

### DIFF
--- a/openxr/Cargo.toml
+++ b/openxr/Cargo.toml
@@ -32,5 +32,40 @@ ctrlc = "3.1.5"
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3", features = ["profileapi"] }
 
+[target.'cfg(target_os = "android")'.dependencies]
+ndk-glue = "0.3"
+
 [package.metadata.docs.rs]
 features = ["linked", "loaded", "mint"]
+
+[[example]]
+name = "vulkan"
+
+[[example]]
+name = "vulkan-android"
+path = "examples/vulkan.rs"
+crate-type = ["cdylib"]
+
+# The following manifest metadata is used by cargo-apk to configure the example Android app for the Oculus Quest 1 and Quest 2.
+# It does not affect the openxr crate.
+[package.metadata.android]
+build_targets = ["aarch64-linux-android"]
+runtime_libs = "examples/libs"
+
+[package.metadata.android.sdk]
+min_sdk_version = 21
+target_sdk_version = 30
+
+[package.metadata.android.application.activity]
+theme = "@android:style/Theme.Black.NoTitleBar.Fullscreen"
+config_changes = "density|keyboard|keyboardHidden|navigation|orientation|screenLayout|screenSize|uiMode"
+launch_mode = "singleTask"
+orientation = "landscape"
+resizeable_activity = "false"
+
+[[package.metadata.android.application.activity.intent_filter]]
+actions = ["android.intent.action.MAIN"]
+categories = [
+    "com.oculus.intent.category.VR",
+    "android.intent.category.LAUNCHER",
+]
--- a/openxr/examples/vulkan.rs
+++ b/openxr/examples/vulkan.rs
@@ -6,6 +6,7 @@
 //! largely decouple its Vulkan and OpenXR components and handle errors gracefully.
 
 use std::{
+    ffi::{CStr, CString},
     io::Cursor,
     sync::{
         atomic::{AtomicBool, Ordering},
@@ -22,7 +23,8 @@ use ash::{
 use openxr as xr;
 
 #[allow(clippy::field_reassign_with_default)] // False positive, might be fixed 1.51
-fn main() {
+#[cfg_attr(target_os = "android", ndk_glue::main)]
+pub fn main() {
     // Handle interrupts gracefully
     let running = Arc::new(AtomicBool::new(true));
     let r = running.clone();
@@ -37,6 +39,9 @@ fn main() {
     let entry = xr::Entry::load()
         .expect("couldn't find the OpenXR loader; try enabling the \"static\" feature");
 
+    #[cfg(target_os = "android")]
+    entry.initialize_android_loader().unwrap();
+
     // OpenXR will fail to initialize if we ask for an extension that OpenXR can't provide! So we
     // need to check all our extensions before initializing OpenXR with them. Note that even if the
     // extension is present, it's still possible you may not be able to use it. For example: the
@@ -47,11 +52,19 @@ fn main() {
     // If a required extension isn't present, you want to ditch out here! It's possible something
     // like your rendering API might not be provided by the active runtime. APIs like OpenGL don't
     // have universal support.
-    assert!(available_extensions.khr_vulkan_enable2);
+    assert!(available_extensions.khr_vulkan_enable || available_extensions.khr_vulkan_enable2);
 
     // Initialize OpenXR with the extensions we've found!
     let mut enabled_extensions = xr::ExtensionSet::default();
-    enabled_extensions.khr_vulkan_enable2 = true;
+    if available_extensions.khr_vulkan_enable2 {
+        enabled_extensions.khr_vulkan_enable2 = true;
+    } else {
+        enabled_extensions.khr_vulkan_enable = true;
+    }
+    #[cfg(target_os = "android")]
+    {
+        enabled_extensions.khr_android_create_instance = true;
+    }
     let xr_instance = entry
         .create_instance(
             &xr::ApplicationInfo {
@@ -90,6 +103,7 @@ fn main() {
     let reqs = xr_instance
         .graphics_requirements::<xr::Vulkan>(system)
         .unwrap();
+
     if vk_target_version_xr < reqs.min_api_version_supported
         || vk_target_version_xr.major() > reqs.max_api_version_supported.major()
     {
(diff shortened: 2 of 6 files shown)


Ralith:
Thanks! An initial concern: as illustrated in the linked reference, don't we also need an `xrInitializeLoaderKHR` call?

As I'm not qualified to review JNI esoterica myself, I'd like success reports from manual testing on android at minimum.

zarik5:
Yes, in the work I'm doing on bevy I'm also sending the same Java pointers to the loader using the extension. It is easy enough to do with the function pointers and it doesn't require extra modifications to openxrs. I could also contribute it directly to openxrs (inside the entry creation call), but I should definitely have something running first (which I'm deferring because I currently don't have the software + hardware ready).

zarik5:
I added `Entry::initialize_android_loader()` and updated the vulkan example for android support.

This PR is now ready for review, with some caveats:

* I tested this on the Oculus Quest 1 and Quest 2 but I'm still waiting reports about the Oculus Go
* It requires cargo-apk from [this PR](https://github.com/rust-windowing/android-ndk-rs/pull/138)
* Due to the lack `khr_vulkan_enable`support, the vulkan example now has much more boilerplate.

openxr/src/entry.rs
```diff
@@ -108,7 +108,31 @@ impl Entry {
         Ok(f.unwrap())
     }
 
-    /// Create an OpenXR instance with certain extensions enabled
+    /// Initialize Android loader. This must be called before `create_instance()`.
+    #[cfg(target_os = "android")]
+    pub fn initialize_android_loader(&self) -> Result<()> {
```
Ralith, line 113: Could we do this automatically at the start of `create_instance`?

openxr/examples/vulkan.rs
```diff
@@ -37,21 +40,32 @@ fn main() {
...
     // If a required extension isn't present, you want to ditch out here! It's possible something
     // like your rendering API might not be provided by the active runtime. APIs like OpenGL don't
     // have universal support.
-    assert!(available_extensions.khr_vulkan_enable2);
+    assert!(available_extensions.khr_vulkan_enable || available_extensions.khr_vulkan_enable2);
 
     // Initialize OpenXR with the extensions we've found!
     let mut enabled_extensions = xr::ExtensionSet::default();
-    enabled_extensions.khr_vulkan_enable2 = true;
+    enabled_extensions.khr_vulkan_enable = available_extensions.khr_vulkan_enable;
```
Ralith, line 62: We should not enable version 1 if version 2 is available. Additionally, a comment should be added documenting that the fallback path is a temporary compatibility measure for Oculus runtimes.

openxr/examples/vulkan.rs
```diff
@@ -37,21 +40,32 @@ fn main() {
...
+
     // OpenXR will fail to initialize if we ask for an extension that OpenXR can't provide! So we
     // need to check all our extensions before initializing OpenXR with them. Note that even if the
     // extension is present, it's still possible you may not be able to use it. For example: the
     // hand tracking extension may be present, but the hand sensor might not be plugged in or turned
     // on. There are often additional checks that should be made before using certain features!
     let available_extensions = entry.enumerate_extensions().unwrap();
 
+    dbg!(&available_extensions);
+
```
Ralith, line 54: Stray debug print?

openxr/examples/vulkan.rs
```diff
@@ -108,26 +143,85 @@ fn main() {
...
+                system,
+                buffer.len() as _,
+                &mut data_len as _,
+                &mut buffer as *mut _ as _,
+            );
+            if res.into_raw() < 0 {
+                panic!("{}", res);
+            }
+            let vk_instance_exts =
+                String::from_utf8_unchecked(buffer[0..data_len as usize - 1].to_vec())
```
Ralith, line 179: This should use `str::from_utf8`.

openxr/examples/vulkan.rs
```diff
@@ -108,26 +143,85 @@ fn main() {
...
+                .exts()
+                .khr_vulkan_enable
+                .unwrap()
+                .get_vulkan_instance_extensions)(
+                xr_instance.as_raw(),
+                system,
+                buffer.len() as _,
+                &mut data_len as _,
+                &mut buffer as *mut _ as _,
+            );
```
Ralith, line 174: This (and the later call for device extensions) should use the proper two-call pattern and a `Vec` buffer to ensure sufficient space is allocated. It's fine to copy the internal helpers to accomplish this (though we might want to consider bundling them into the `sys` crate).

Ralith (changes requested):
Thanks, left some comments. Please split the changes to the example out into a separate PR (or, if you're comfortable with history rewrites, maintain them in a separate commit) so that they can be easily reverted once Oculus updates their runtimes.

openxr/Cargo.toml
```diff
@@ -37,3 +37,38 @@ ndk-glue = "0.3"
...
+orientation = "landscape"
+resizeable_activity = "false"
+
+[[package.metadata.android.intent_filter]]
+actions = ["android.intent.action.MAIN"]
+categories = [
+    "com.oculus.intent.category.VR",
+    "android.intent.category.LAUNCHER",
+    "android.intent.category.INFO",
+]
```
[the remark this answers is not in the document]
zarik5: This metadata applies to all `cdylib` crates and examples. Since `openxr` and `openxr-sys`are `lib`crates, they are unaffected.

openxr/src/entry.rs
zarik5: It would make more sense to insert this call inside `Entry::load()`, the problem is that they return different error types. While not totally correct, it is possible to insert this inside `create_instance`, and I also verified that calling this twice does not have any adverse impact.

openxr/examples/vulkan.rs
zarik5: i'll remove it

openxr/examples/vulkan.rs
zarik5: I'll remove this code when I add proper support for `khr_vulkan_enable`.

openxr/examples/vulkan.rs
zarik5: I'll remove this code when I add proper support for `khr_vulkan_enable`.

openxr/src/entry.rs
Ralith: Hmm, I'd rather not rely on coincidence there; perhaps keeping it separate is the most practical option.

openxr/Cargo.toml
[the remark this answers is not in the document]
Ralith: Makes sense. Please add a comment to the Cargo.toml explaining this. Might also be helpful to reference it from examples/README.md as a model for how to set up a Quest app.

zarik5:
I'm going to add here commits related to the implementation of the `khr_vulkan_enable` wrapper, for the sake of testing. I'll rebase later.

zarik5:
The Oculus Go is probably compatible with OpenXR and this crate (I'm waiting for some last reports). The problem is that it supports only Vulkan 1.0 so the example doesn't run without heavy modifications, so I'm not listing it as compatible (but it will be compatible on bevy).

Ralith:
It might not be too bad to refactor the example to use the multiview extension explicitly assuming the Go implements it, but I'm just as happy to keep things more concise/modern.

zarik5:
@Ralith The Oculus Go supports the extension `VK_KHR_multiview`, but adding code to support it will bloat the example even more, where it is more important showing the openxr crate usage.

zarik5:
I'm ok with keeping the example on Vulkan 1.1 (without Oculus Go support). Another argument is that the cargo-apk metadata was a mashup of old and new entries for the sake of the Oculus Go support, when the user would probably decide to maintain separate Cargo.toml files.

Ralith (approved):
Thanks!

openxr/examples/vulkan.rs
```diff
@@ -47,11 +52,19 @@ fn main() {
...
-    enabled_extensions.khr_vulkan_enable2 = true;
+    if available_extensions.khr_vulkan_enable2 {
+        enabled_extensions.khr_vulkan_enable2 = true;
+    } else {
+        enabled_extensions.khr_vulkan_enable = true;
+    }
+    #[cfg(target_os = "android")]
+    {
+        enabled_extensions.khr_android_create_instance = true;
+    }
```
Ralith, line 67: Style nit, since more work is needed anyway:
```suggestion
    enabled_extensions.khr_android_create_instance = cfg!(target_os = "android");
```

openxr/Cargo.toml
```diff
@@ -32,5 +32,40 @@ ctrlc = "3.1.5"
...
 [package.metadata.docs.rs]
 features = ["linked", "loaded", "mint"]
+
+[[example]]
+name = "vulkan"
+
+[[example]]
+name = "vulkan-android"
+path = "examples/vulkan.rs"
+crate-type = ["cdylib"]
```
Ralith, line 47: Looks like this is causing lint issues on non-android. Can we guard the example to android-only somehow? Failing that, we should move most of the code into a common module, and add suitable conditional lint `allow`s in the vulkan-android-specific source file, perhaps combined with a `panic` with a suitable message if ran on non-android.

Ralith (changes requested):

openxr/Cargo.toml
zarik5: I'll look into it.

openxr/examples/vulkan.rs
zarik5: This doesn't work because `khr_android_create_instance` field itself is cfg-gated to Android only.

openxr/Cargo.toml
zarik5: The solution was to add also the "bin" crate-type. I've seen doing this elsewhere but I wasn't aware it fixes the lints.

openxr/examples/vulkan.rs
Ralith: Oh, right.

openxr/Cargo.toml
Ralith: Huh. Does this result a useless binary being produced as well? Is the android build process unaffected?

zarik5: I just tried, the binary gets created in the target folder, but it doesn't get used. it doesn't seem to affect compilation times.

zarik5: I added a commit to use `#[allow(unused)]` instead. This is something that can be integrated in ndk_glue.

zarik5: I found a more elegant solution, make the main `pub`.

zarik5:
I fixed the examples README with the crates.io link for cargo-apk. I think now it's all ready for merging.

Ralith (approved):

Issue #84, opened 2021-06-28
Vulkan with `khr_vulkan_enable` instead of `khr_vulkan_enable2`


Tenebryo:
I was trying to get the Vulkan example to work, but the headset I'm using only supports `khr_vulkan_enable` and not `khr_vulkan_enable2`. How would I have to change the example to accommodate this? I'm not sure how much the specific device matters, but I'm trying to run the example on the quest 2 over USB via oculus link.

Ralith:
High-level bindings are not available for the obsolete Vulkan extension; your vendor (Oculus) should implement support for the current one. In the mean time, you could:

- Use version 0.13.1 of this crate, the last release with high-level support for the old extension, or
- Call the raw functions from the old extension yourself via e.g. `Instance::fp`.

0.13.1 isn't too old yet, so that's probably the easiest thing for now, and saves you some work if Oculus catches up with everyone else before too much longer.

Tenebryo:
Thanks! I'll probably use version 0.13.1 for now, and otherwise use it as a reference if I need to go with the low-level approach later.

zarik5:
I'm having the same problem running the vulkan example natively on the Quest 2. I think Oculus has no incentive to upgrade to `khr_vulkan_enable2` because 1) `khr_vulkan_enable` it is part of the OpenXR 1.0 standard 2) mobile SoCs only have one GPU and the problem that `khr_vulkan_enable2` solves is not even there (the application cannot select the wrong adapter). For these reasons I think it is reasonable to re-add `khr_vulkan_enable` support to the higher-level API.

Ralith:
> I think Oculus has no incentive to upgrade

Oculus's incentive to update to the revised extension is for compatibility with applications, same as their incentive for implementing OpenXR at all. I recommend reaching out to them if you're concerned about whether they have plans to do so.

> 1) khr_vulkan_enable it is part of the OpenXR 1.0 standard

It's an extension, not part of 1.0 core.

> mobile SoCs only have one GPU and the problem that khr_vulkan_enable2 solves is not even there (the application cannot select the wrong adapter)

The problem that the revised extension solves is allowing the runtime to set features and supply extension structures itself; it has nothing to do with selecting between multiple adapters, which was already accounted for in the original extension.

zarik5:
> The problem that the revised extension solves is allowing the runtime to set features and supply extension structures itself; it has nothing to do with selecting between multiple adapters, which was already accounted for in the original extension.

[...]

PR #85, opened 2021-07-05
Add back khr_vulkan_enable support

zarik5:
As proposed in #84.

The code is mostly copy-pasted from the tag v0.13.1

Tested on the Oculus Quest 2.

### DIFF
--- a/openxr/src/instance.rs
+++ b/openxr/src/instance.rs
@@ -229,10 +229,10 @@ impl Instance {
     ///
     /// Returns a space-delimited list of Vulkan instance extension names.
     ///
-    /// **Note:** This method requires `khr_vulkan_enable` extension. Applications should use
+    /// **Note:** This method requires the `khr_vulkan_enable` extension. Applications should use
     /// `khr_vulkan_enable2` instead, if possible.
     #[inline]
-    pub fn vulkan_instance_extensions(&self, system: SystemId) -> Result<String> {
+    pub fn vulkan_legacy_instance_extensions(&self, system: SystemId) -> Result<String> {
         get_str(|input, output, buf| unsafe {
             (self.vulkan_legacy().get_vulkan_instance_extensions)(
                 self.as_raw(),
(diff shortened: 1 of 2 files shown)


openxr/src/instance.rs
```diff
@@ -225,11 +225,48 @@ impl Instance {
...
-    /// Get a suitable [`VkPhysicalDevice`] for use with a particular `system`
+    /// Identify the Vulkan instance extensions required by a system
+    ///
+    /// Returns a space-delimited list of Vulkan instance extension names.
     ///
-    /// Call [`Instance::create_vulkan_instance()`] first to get a suitable `VkInstance`.
+    /// **Note:** This method requires `khr_vulkan_enable` extension. Applications should use
+    /// `khr_vulkan_enable2` instead, if possible.
+    #[inline]
+    pub fn vulkan_instance_extensions(&self, system: SystemId) -> Result<String> {
```
Ralith, line 235:
```suggestion
    pub fn vulkan_legacy_instance_extensions(&self, system: SystemId) -> Result<String> {
```

openxr/src/instance.rs
```diff
@@ -225,11 +225,48 @@ impl Instance {
...
+    }
+
+    /// Identify the Vulkan device extensions required by a system
+    ///
+    /// Returns a space-delimited list of Vulkan device extension names.
+    ///
+    /// **Note:** This method requires `khr_vulkan_enable` extension. Applications should use
+    /// `khr_vulkan_enable2` instead, if possible.
+    #[inline]
+    pub fn vulkan_device_extensions(&self, system: SystemId) -> Result<String> {
```
Ralith, line 254:
```suggestion
    pub fn vulkan_legacy_device_extensions(&self, system: SystemId) -> Result<String> {
```

openxr/src/instance.rs
```diff
@@ -225,11 +225,48 @@ impl Instance {
         Ok(Ok(instance))
     }
 
-    /// Get a suitable [`VkPhysicalDevice`] for use with a particular `system`
+    /// Identify the Vulkan instance extensions required by a system
+    ///
+    /// Returns a space-delimited list of Vulkan instance extension names.
     ///
-    /// Call [`Instance::create_vulkan_instance()`] first to get a suitable `VkInstance`.
+    /// **Note:** This method requires `khr_vulkan_enable` extension. Applications should use
```
Ralith, line 232:
```suggestion
    /// **Note:** This method requires the `khr_vulkan_enable` extension. Applications should use
```

openxr/src/instance.rs
```diff
@@ -225,11 +225,48 @@ impl Instance {
...
+                buf,
+            )
+        })
+    }
+
+    /// Identify the Vulkan device extensions required by a system
+    ///
+    /// Returns a space-delimited list of Vulkan device extension names.
+    ///
+    /// **Note:** This method requires `khr_vulkan_enable` extension. Applications should use
```
Ralith, line 251:
```suggestion
    /// **Note:** This method requires the `khr_vulkan_enable` extension. Applications should use
```

Ralith (changes requested):
Wow, that's way smaller a change than I expected. Awesome!

Let's name the legacy methods a little more distinctively to avoid confusion with the new ones.

zarik5:
The PR should be ready for merging.

Ralith (approved):
Thanks!

PR #91, opened 2021-10-13
Derives `Debug` for most sys structs

ashpil:
This derives `Debug` for most sys structs, aiding in debugging.

There was some concern about whether this increases compilation time - I found only about a 5% increase in my benchmarks, making this PR so you can try yourself.

### DIFF
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -1971,7 +1971,9 @@ impl StructMeta {
 
 impl std::ops::BitOrAssign for StructMeta {
     fn bitor_assign(&mut self, rhs: Self) {
+        self.has_unprintable |= rhs.has_unprintable;
         self.has_pointer |= rhs.has_pointer;
+        self.has_array |= rhs.has_array;
         self.has_graphics |= rhs.has_graphics;
     }
 }


generator/src/main.rs
```diff
@@ -1926,6 +1929,7 @@ impl Parser {
 
 #[derive(Debug, Copy, Clone, Default)]
 struct StructMeta {
+    has_function_pointer: bool,
```
ashpil, line 1932: Not sure if there's a better way to do this than introducing this field

generator/src/main.rs
```diff
@@ -1576,6 +1578,7 @@ impl Parser {
     fn compute_meta(&self, name: &str, s: &Struct) -> StructMeta {
         let mut out = StructMeta::default();
         for member in &s.members {
+            out.has_function_pointer |= member.ty.starts_with("PFN") || member.ty == "LUID";
```
Ralith, line 1581: I don't think `LUID`s are function pointers. Maybe rename to `has_unprintable`?

generator/src/main.rs
Ralith: Short of generating a `Debug` impl explicitly for these cases (which I don't particularly care about), I think it's the most pragmatic solution.

Ralith (commented):
Thanks! On my laptop, this increases the compiletime for sys by ~0.4s or 25%, and the high level crate by ~0.2s or 7%. That's not nothing, but given that it's small in absolute terms, dwarfed by the safety/ergonomic abstractions, and should parallelize well with other parts of downstream crates' dep graphs, I think it's acceptable.

generator/src/main.rs
ashpil: Done.

ashpil:
Also noticed `BitOrAssign` for `StructMeta` doesn't update `has_array` - not sure if you specifically excluded certain fields from that, so I didn't add `has_unprintable` either.

Ralith:
I think that was unintended; they should both be propagated, though I doubt this will change the output.

ashpil:
@Ralith so is there anything more this PR needs?

PR #97, opened 2021-12-14
update for openxr 1.0.20

kcking:
I took a crack at updating the generator for openxr 1.0.20. A couple things I found relatively general solutions to, and a couple I opted for a quick one-off change. Here's a list of things I changed

- allow lengths that don't end in `.`
- handle naming collision between `XR_SWAPCHAIN_USAGE_INPUT_ATTACHMENT_BIT_{MND,KHR}`
- manually import `XrViveTrackerPathsHTCX`
- handle `Action` setter parameter which has generic `ActionTy`, akin to `G: Graphics` type parameter
- handle BaseHeader setter arguments
- handle `_COUNT` and `_SIZE_FB` length suffixes
- add a manual (but trivial) `FoveationProfileFB` wrapper struct

Let me know if this type of contribution is desirable and if there's anything you'd like to be approached differently!

### DIFF
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -981,6 +981,14 @@ impl Parser {
                 let value = match v.value {
                     ConstantValue::Literal(x) => quote! {Self(1 << #x)},
                     ConstantValue::Alias(ref x) => {
+                        // avoids collision between XR_SWAPCHAIN_USAGE_INPUT_ATTACHMENT_BIT_{MND,KHR}
+                        if bitmask
+                            .values
+                            .iter()
+                            .any(|v| value_name == xr_bitmask_value_name(name, &v.name))
+                        {
+                            return quote! {};
+                        }
                         let ident = xr_enum_value_name(name, x);
                         quote! { Self::#ident }
                     }
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -1916,7 +1916,7 @@ impl Parser {
                         ViveTrackerPathsHTCX
                     },
                     quote! {
-                        unsafe { (self.0).#ident.as_ref() }.unwrap().into()
+                        (*unsafe { self.0.#ident.as_ref() }.unwrap()).into()
                     },
                 )
             } else {
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -1722,6 +1722,8 @@ impl Parser {
                 type_args
             } else if m.ty == "XrSwapchain" || m.ty == "XrSession" {
                 quote! { <G> }
+            } else if m.ty == "XrAction" {
+                quote! { <ATY> }
             } else {
                 quote! {}
             };
@@ -1779,9 +1781,14 @@ impl Parser {
                     }
                 }
             };
+
+            let fn_type_params = match &m.ty[..] {
+                "XrAction" => quote! { <ATY: ActionTy> },
+                _ => quote! {},
+            };
             Some(quote! {
                 #[inline]
-                pub fn #ident(mut self, value: #ty) -> Self {
+                pub fn #ident#fn_type_params(mut self, value: #ty) -> Self {
                     #init
                     self
                 }
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -1916,7 +1916,7 @@ impl Parser {
                         ViveTrackerPathsHTCX
                     },
                     quote! {
-                        (self.0).#ident.into()
+                        unsafe { (self.0).#ident.as_ref() }.unwrap().into()
                     },
                 )
             } else {
@@ -1930,13 +1930,18 @@ impl Parser {
             })
         });
 
+        let sys_raw_ident_str = format!("[sys::{}]", raw_ident);
         quote! {
             #[derive(Copy, Clone)]
             pub struct #ident<'a>(&'a sys::#raw_ident);
 
             impl<'a> #ident<'a> {
                 #[inline]
-                pub fn new(inner: &'a sys::#raw_ident) -> Self {
+                /// # Safety
+                /// `inner` must be valid event data according to the OpenXR spec. Refer to
+                #[doc = #sys_raw_ident_str]
+                /// for more information.
+                pub unsafe fn new(inner: &'a sys::#raw_ident) -> Self {
                     Self(inner)
                 }
 
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -918,8 +918,7 @@ impl Parser {
                         |x| {
                             let mut reason = x.to_string();
                             reason.get_mut(0..1).unwrap().make_ascii_lowercase();
-                            assert!(reason.ends_with('.'));
-                            reason.truncate(reason.len() - 1);
+                            reason = reason.trim_end_matches(".").to_string();
                             reason
                         },
                     );
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -2138,7 +2138,9 @@ fn xr_var_ty(member: &Member) -> TokenStream {
         if let Ok(len) = len.parse::<usize>() {
             quote! { [#ty; #len] }
         } else {
-            assert!(len.starts_with("XR_MAX_"));
+            assert!(
+                len.starts_with("XR_MAX_") || len.ends_with("_COUNT") || len.ends_with("_SIZE_FB")
+            );
             let len = Ident::new(&len[3..], Span::call_site());
             quote! { [#ty; #len] }
         }
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -507,7 +507,7 @@ impl Parser {
                     self.finish_element();
                 }
                 Characters(ch) => {
-                    if define_ty.is_some() {
+                    if define_ty.is_some() || define_name.is_some() {
                         define_val = Some(ch);
                     }
                 }
@@ -520,6 +520,14 @@ impl Parser {
                 _ => {}
             }
         }
+        match (&define_name, &define_val) {
+            (Some(name), Some(val)) => {
+                if let Ok(val) = val.parse::<usize>() {
+                    self.api_constants.push((name.into(), val))
+                }
+            }
+            _ => {}
+        };
         if define_name.as_ref().map(|x| &x[..]) == Some("XR_CURRENT_API_VERSION") {
             let version = define_val.unwrap();
             assert!(version.starts_with('('));
@@ -1182,7 +1190,7 @@ impl Parser {
         quote! {
             //! Automatically generated code; do not edit!
 
-            #![allow(non_upper_case_globals, clippy::unreadable_literal, clippy::identity_op)]
+            #![allow(non_upper_case_globals, clippy::unreadable_literal, clippy::identity_op, unused)]
             use std::fmt;
             use std::mem::MaybeUninit;
             use std::os::raw::{c_void, c_char};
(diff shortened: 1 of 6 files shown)


openxr/src/generated.rs
```diff
@@ -1359,6 +1620,54 @@ impl<'a> DisplayRefreshRateChangedFB<'a> {
...
+pub struct ViveTrackerConnectedHTCX<'a>(&'a sys::EventDataViveTrackerConnectedHTCX);
+impl<'a> ViveTrackerConnectedHTCX<'a> {
+    #[inline]
+    pub fn new(inner: &'a sys::EventDataViveTrackerConnectedHTCX) -> Self {
+        Self(inner)
+    }
+    #[inline]
+    pub fn paths(self) -> *mut ViveTrackerPathsHTCX {
+        (self.0).paths
+    }
```
Ralith, line 1645: This isn't very ergonomic/safe. How about a wrapper iterator or something?

openxr/src/foveation_profile_fb.rs
```diff
@@ -0,0 +1,21 @@
+use std::sync::Arc;
+
+use crate::*;
+
+pub struct FoveationProfileFB {
```
Ralith, line 5: How would this be used?

openxr/src/generated.rs
```diff
@@ -3515,6 +4196,92 @@ pub(crate) mod builder {
...
+        #[inline]
+        pub fn into_raw(self) -> sys::InteractionProfileAnalogThresholdVALVE {
+            self.inner
+        }
+        #[inline]
+        pub fn as_raw(&self) -> &sys::InteractionProfileAnalogThresholdVALVE {
+            &self.inner
+        }
+        #[inline]
+        pub fn action(mut self, value: &'a Action<ATY>) -> Self {
```
Ralith, line 4242: Could the type variable be pushed down to this setter, rather than associated with the entire struct?

Ralith (changes requested):
Thanks, this is very welcome! For ease of review, could you reformat this PR as a series of minimal, atomic commits that each address a single thing?

kcking:
> Thanks, this is very welcome! For ease of review, could you reformat this PR as a series of minimal, atomic commits that each address a single thing?

Thanks for taking a look! I'm getting the commits tidied up and then will respond to your comments

generator/src/main.rs
```diff
@@ -1898,11 +1898,29 @@ impl Parser {
...
             let ident = xr_var_name(&m.name);
+            // dbg!(self.structs.keys());
+            // dbg!(&m.ty);
             let (ty, value) = if m.ty == "XrBool32" {
                 (quote! { bool }, quote! { (self.0).#ident.into() })
             } else if self.handles.contains(&m.ty) {
                 let ty = xr_var_ty(m);
                 (quote! { sys::#ty }, quote! { (self.0).#ident })
+            } else if self.is_ptr_chain(&m.ty) {
+                let ty = xr_var_ty(m);
```
kcking, line 1907: @Ralith I'm having trouble replying to the original comment, but I wrapped the pointer chain reader in a Vec in response to your recommendation to expose an iterator.

This was simpler than generating code for a wrapper Iterator type. Let me know if this is what you had in mind!

openxr/src/foveation_profile_fb.rs
Ralith: If we're offering high-level bindings of this extension, it feels odd to have no way to actually *get* a `FoveationProfileFB` value. Also, shouldn't we handle destruction as well?

generator/src/main.rs
Ralith: This produces a `Vec<*mut ViveTrackerPathsHTCX>`, which still requires the caller to write unsafe code to deref raw pointers. An ideal solution would return `impl Iterator<Item=ViveTrackerPathsHTCX>`, where `ViveTrackerPathsHTCX` is a handwritten type that omits the `s_type` and `pNext` members. It probably makes sense to write most, or even all, of the necessary code outside of the generator in a utility type/function, similar to the existing helpers in openxr/src/lib.rs.

Ralith: Don't worry too much about trying to implement a general solution if it becomes difficult; we don't know how often this pattern might come up in the future.

kcking: I just realized that in [the spec](https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XrViveTrackerPathsHTCX) `XrViveTrackerPathsHTCX.next` says
> next is NULL or a pointer to the next structure in a structure chain. No such structures are defined in core OpenXR or this extension.

I suppose this means the struct chain will only ever have one element. Does it make sense to just expose it as a single wrapper struct like below then?
```
pub struct ViveTrackerPaths {
    pub persistent: Path,
    pub role: Path,
}
```

Ralith: Oh, good, that's much simpler then. That would be ideal, yeah!

openxr/src/foveation_profile_fb.rs
kcking: Good points! I added a way to create a profile, and added a `destroy` method. I opted against a destructor because 
1. we need access to the destroy_profile function pointer
2. the struct may have come from a raw handle in which case we dont know if the caller intends to continue using it after this wrapper struct is dropped.
let me know what you think!

generator/src/main.rs
kcking: added a manual wrapper in https://github.com/Ralith/openxrs/pull/97/commits/b287335bbdd253312d1db20faf119e445f80a997 , lmk what you think!

openxr/src/vive_tracker_paths.rs
```diff
@@ -0,0 +1,20 @@
+use crate::Path;
+
+pub struct ViveTrackerPathsHTCX {
+    pub persistent: Path,
+    pub role: Option<Path>,
+}
+
+impl From<*mut sys::ViveTrackerPathsHTCX> for ViveTrackerPathsHTCX {
+    fn from(paths: *mut sys::ViveTrackerPathsHTCX) -> Self {
+        let paths = unsafe { *paths };
```
Ralith, line 10: This is unsound; safe code could pass in a null pointer. Perform the deref in the generated code instead.

kcking: Ah good point. Couldn't a user build the wrapper type themselves with a null `paths` anyways? Should we change the return type to `Option` and return `None` when it's null instead?

openxr/src/foveation_profile_fb.rs
```diff
@@ -0,0 +1,76 @@
...
+    ) -> Result<Self> {
+        let instance = session.instance();
+        let entry = instance.entry();
+        let fp = raw::FoveationFB::load(entry, instance.as_raw())?;
+
+        let mut level_profile = level_profile.map(|lp| {
+            let mut c = sys::FoveationLevelProfileCreateInfoFB::out(std::ptr::null_mut());
+            (*c.as_mut_ptr()).vertical_offset = lp.vertical_offset;
+            (*c.as_mut_ptr()).level = lp.level;
+            (*c.as_mut_ptr()).dynamic = lp.dynamic;
```
Ralith, line 28: Each of these is unsound, because they construct an `&mut T` addressing partially uninitialized data. The `out` method is intended for OpenXR *output* structures, i.e. structures that OpenXR populates; for structures we populate, use conventional Rust structure construction syntax.

openxr/src/foveation_profile_fb.rs
```diff
@@ -0,0 +1,76 @@
...
+            c.assume_init()
+        });
+        let next = if let Some(level_profile) = level_profile.as_mut() {
+            std::mem::transmute::<_, &mut BaseOutStructure>(level_profile)
+        } else {
+            std::ptr::null_mut()
+        };
```
Ralith, line 35: Favor `cast`ing pointers over transmuting references.

openxr/src/foveation_profile_fb.rs
```diff
@@ -0,0 +1,76 @@
...
+
+        let mut create_info = sys::FoveationProfileCreateInfoFB::out(next).assume_init();
```
Ralith, line 37: This is unsound because it assumes a partially initialized value is fully initialized. As above, use conventional Rust structure construction syntax instead.

openxr/src/foveation_profile_fb.rs
```diff
@@ -0,0 +1,76 @@
...
+        let res = (fp.destroy_foveation_profile)(self.inner.0);
+        if res.into_raw() < 0 {
+            return Err(res);
+        }
+
+        Ok(())
+    }
+
+    #[inline]
+    pub unsafe fn from_raw(handle: sys::FoveationProfileFB) -> Self {
```
Ralith, line 64: This should have a doc comment similar to those present on other `from_raw` methods in the crate.

Ralith (changes requested):

openxr/src/vive_tracker_paths.rs
Ralith: I'm not sure what you mean. If the user derefs a null pointer, they've invoked UB. We should not provide safe interfaces that permit this to occur. Null `Path`s are not null pointers, and are fortunately much safer to handle.

kcking: I'm specifically talking about [this field](https://github.com/Ralith/openxrs/pull/97/files#diff-2d9ed58d2f856284b7dc54da37b3e9690c24a83d3d15358961e0a2cce8bdea96R5358) being a null pointer. In that case, wouldn't a deref in the [generated code](https://github.com/Ralith/openxrs/pull/97/files#diff-9302c9ce141dbd6e8d9a2fa27588e9701a01ec8f3eb53b52e1ccb9676cb3a189R1644) still be UB?

Ralith: Neither of those links address a specific line of code, so I'm not sure what you're referring to.

kcking: Oh 🤦 looks like links to lines in review diffs don't work.

I'm specifically talking about [this field](https://github.com/Ralith/openxrs/blob/b287335bbdd253312d1db20faf119e445f80a997/sys/src/generated.rs#L5358) being a null pointer. In that case, wouldn't a deref in the [generated code](https://github.com/Ralith/openxrs/blob/b287335bbdd253312d1db20faf119e445f80a997/openxr/src/generated.rs#L1644) still be UB?

Ralith: Ah, yes. Is this really the first time we've needed to traverse a pointer in an event? Huh. We should probably make `ViveTrackerConnectedHTCX::new` unsafe, with the requirement that the argument must satisfy the requirements of the OpenXR spec placed upon such values when delivered by OpenXR.

kcking:
Thank you for the super helpful comments! I believe I followed all of your suggestions. 

I made all EventData constructors unsafe as you mentioned. I generated a doc comment to point the user toward the analagous struct in `sys` which has a pointer to the openxr spec.

Also I added the `Session::create_foveation_profile` impl in `foveation_profile_fb.rs` instead of `session.rs` since it seemed more relevant.

Ralith (changes requested):
Thanks, this looks great! Just a couple minor issues left.

kcking:
This latest commit should address both issues since they affected the same line :)

Ralith (approved):
Thanks!

kcking:
Ah I can fix those lints

Ralith:
Released in 0.15.5.

kcking:
> Released in 0.15.5.

🎉 thank you for all of the help with this!

zarik5:
The foveation profile bindings seem to be incomplete. You need also bindings for `xrUpdateSwapchainFB` in `XR_FB_swapchain_update_state` to be able to apply the foveation profile.

PR #104, opened 2022-02-08
Add `Debug` impl for `instance::SystemProperties` and `vk::vulkan::Requirements`

zjijz:
I'm not quite sure why this wasn't already included. `SystemId`, `SystemGraphicsProperties`, and `SystemTrackingProperties` are both `Debug`, and the other properties are stdlib types.

Does this look okay?


Ralith:
LGTM, thanks! This just hadn't come up yet.

PR #107, opened 2022-02-12
Update sys crate to use OpenXR 1.0.22

zjijz:
This does the work of getting the `openxr-sys` crate to use the headers from OpenXR SDK 1.0.22. Full OpenXR changelog is here: https://github.com/KhronosGroup/OpenXR-SDK/releases/tag/release-1.0.22.

There will be a follow-on PR that deals with actually updating the higher-level `openxr` crate.

### DIFF
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -47,7 +47,6 @@ struct Parser {
     structs: IndexMap<String, Struct>,
     struct_aliases: Vec<(String, String)>,
     api_constants: Vec<(String, usize)>,
-    // TODO: Bind output generation to look at api_aliases.
     api_aliases: Vec<(String, String)>,
     commands: IndexMap<String, Command>,
     cmd_aliases: Vec<(String, String)>,
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -47,6 +47,8 @@ struct Parser {
     structs: IndexMap<String, Struct>,
     struct_aliases: Vec<(String, String)>,
     api_constants: Vec<(String, usize)>,
+    // TODO: Bind output generation to look at api_aliases.
+    api_aliases: Vec<(String, String)>,
     commands: IndexMap<String, Command>,
     cmd_aliases: Vec<(String, String)>,
     extensions: IndexMap<String, Tag>,
@@ -69,6 +71,7 @@ impl Parser {
             structs: IndexMap::new(),
             struct_aliases: Vec::new(),
             api_constants: Vec::new(),
+            api_aliases: Vec::new(),
             commands: IndexMap::new(),
             cmd_aliases: Vec::new(),
             extensions: IndexMap::new(),
@@ -267,6 +270,8 @@ impl Parser {
                                 } else {
                                     eprintln!("extension to unrecognized type {}", extends);
                                 }
+                            } else if let Some(alias) = attr(&attributes, "alias") {
+                                self.api_aliases.push((name.into(), alias.into()));
                             } else if name.ends_with("SPEC_VERSION") {
                                 let value = attr(&attributes, "value").unwrap();
                                 ext_version = Some(value.parse().unwrap());
@@ -274,7 +279,11 @@ impl Parser {
                                 let value = attr(&attributes, "value").unwrap();
                                 assert_eq!(&ext_name[..], &value[1..value.len() - 1]);
                             } else {
-                                let value = attr(&attributes, "value").unwrap();
+                                let value = attr(&attributes, "value");
+                                if let None = value {
+                                    println!("{:#?}\n{:#?}", name, attributes);
+                                }
+                                let value = value.unwrap();
                                 self.api_constants
                                     .push((name.into(), value.parse().unwrap()));
                             }
@@ -878,6 +887,14 @@ impl Parser {
             }
         });
 
+        let consts_aliases = self.api_aliases.iter().map(|(name, value)| {
+            let alias_ident = Ident::new(&name[3..], Span::call_site());
+            let og_ident = Ident::new(&value[3..], Span::call_site());
+            quote! {
+                pub const #alias_ident: usize = #og_ident;
+            }
+        });
+
         let enums = self.enums.iter().map(|(name, e)| {
             let ident = xr_ty_name(name);
             let values = e.values.iter().map(|v| {
@@ -1200,6 +1217,7 @@ impl Parser {
             pub const CURRENT_API_VERSION: Version = Version::new(#major, #minor, #patch);
 
             #(#consts)*
+            #(#consts_aliases)*
             #(#enums)*
             #(#bitmasks)*
             #(#handles)*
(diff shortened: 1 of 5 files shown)


zjijz:
@Ralith I saw that a lot of the extensions do not expose a higher-level wrapper in `openxr` and I imagine users just put the `sys` versions of structs in the `next` pointers for specific APIs. Do you see anything in the new update that you think would benefit from a higher-level wrapping?

sys/README.md
```diff
@@ -0,0 +1,21 @@
+# openxr-sys
+
+## Cloning
+
+If you can, clone the repo using `git clone --recurse-submodules`. 
+
+If you already cloned the repo without, you must run (from anywhere in the repo):
+1. `git submodules init`.
+2. `git submodules update`.
```
Ralith, line 9: `git submodule update --init`

sys/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+1. Update the OpenSR-SDK git submodule to a new commit hash or tag:
+  a. `cd openxrs/sys/OpenXR-SDK`.
+  b. `git up <tag/xxx or hash>`.
+  c. `cd ../..` to get back to the `openxrs` repo root and check that the 
+  submodule is updated (should be a single line change to what looks like a file
+  at `openxrs/sys/OpenXR-SDK`).
+2. `cd generator; cargo run --bin generator` to regenerate the `sys` crate.
+3. `cd ..` to go back to the `openxrs` repo root`.
+3. `cargo build && cargo test && cargo fmt` to find any issues that need fixing.
```
Ralith, line 21: `cargo run --bin generator` works from the root, and `cargo fmt` should be ran *before* `cargo build`.

generator/src/main.rs
```diff
@@ -878,6 +887,14 @@ impl Parser {
...
         });
 
+        let consts_aliases = self.api_aliases.iter().map(|(name, value)| {
+            let alias_ident = Ident::new(&name[3..], Span::call_site());
+            let og_ident = Ident::new(&value[3..], Span::call_site());
+            quote! {
+                pub const #alias_ident: usize = #og_ident;
+            }
+        });
+
```
Ralith, line 892: AFAICT these aliases exist only for backwards compatibility. Because we can cleanly break compat, we shouldn't bother generating them.

README.md
```diff
@@ -45,3 +45,5 @@ example, symbols are named according to Rust conventions, enums and
 bitmasks are strongly typed, and many types have helpful `Debug`
 impls. This crate is almost entirely generated from the Khronos XML
 registry.
+
+The bindings are based off OpenXR SDK v1.0.22.
```
Ralith: I'm not sure what the value of documenting this is; it's already implicit in the submodule revision, and it's very likely to get missed in future updates.

Ralith (changes requested):
Thanks!

Ralith:
> Do you see anything in the new update that you think would benefit from a higher-level wrapping?

Generally I wrap things as needed. If nobody needs something, then wrapping is wasted effort.

generator/src/main.rs
zjijz: Should we break backward compatibility though? OpenXR itself tries to preserve it so maybe we should as well (at least for `sys`)?

sys/README.md
zjijz: I couldn't get it to run from the `openxrs` repo root due to lines 34-35 of `generatror/src/main.rs`:
```
let mut sys_out = File::create("../sys/src/generated.rs").unwrap();
let mut hl_out = File::create("../openxr/src/generated.rs").unwrap();
 ```

These seem to assume PWD directory is always one of the children of `openxrs` repo root. 

It looks like line 27 of `generatror/src/main.rs` might have the solution by starting the path at `env!(CARGO_MANIFEST_DIR)` and then moving up based on that fixed path.

File: generator/src/main.rs (hunk 3 above)
Comment (Ralith) [reply generator/src/main.rs:892]: The OpenXR C headers operate under different constraints. We should focus on providing high quality support for the latest version.

File: sys/README.md (hunk 2 above)
Comment (Ralith) [reply sys/README.md:21]: Ah, good catch. No need to fix that in this PR. `cargo fmt` should still be ran before building though.

File: generator/src/main.rs (hunk 3 above)
Comment (zjijz) [reply generator/src/main.rs]: Okay, in FB's infinite wisdom, they didn't update the constants when used in their hand tracking capsule extension, so we're generating the original names:
```
pub struct HandCapsuleFB {
    pub points: [Vector3f; FB_HAND_TRACKING_CAPSULE_POINT_COUNT],
    pub radius: f32,
    pub joint: HandJointEXT,
}
```
and
```
pub struct HandTrackingCapsulesStateFB {
    pub ty: StructureType,
    pub next: *mut c_void,
    pub capsules: [HandCapsuleFB; FB_HAND_TRACKING_CAPSULE_COUNT],
}
```
(ass opposed to the "correct" `HAND_TRACKING_CAPSULE_POINT_COUNT_FB` and `HAND_TRACKING_CAPSULE_COUNT_FB`.

Do you want me to keep the alias constants generating in Rust or try to do a lookup when emitting these structures and try to fix it with the alias map?

Comment (Ralith) [reply generator/src/main.rs:892]: Oh, that's annoying, and should maybe be raised upstream. Ideally I would prefer to work around this in the generator rather than letting it influence our output.

File: sys/README.md [hunk 5]
```diff
@@ -0,0 +1,19 @@
+# openxr-sys
+
+## Cloning
+
+If you can, clone the repo using `git clone --recurse-submodules`. 
+
+If you already cloned the repo without, you must run (from anywhere in the repo): `git submodules update --init`.
```
Ralith, line 7:
```suggestion
If you already cloned the repo without, you must run (from anywhere in the repo): `git submodule update --init`.
```

generator/src/main.rs
```diff
@@ -2171,12 +2181,17 @@ fn xr_var_ty(member: &Member) -> TokenStream {
...
+            Some((_, og)) => og,
+            _ => len,
+        };
         if let Ok(len) = len.parse::<usize>() {
             quote! { [#ty; #len] }
         } else {
-            assert!(
-                len.starts_with("XR_MAX_") || len.ends_with("_COUNT") || len.ends_with("_SIZE_FB")
-            );
+            assert!(len.starts_with("XR_"));
```
zjijz, line 2194: @Ralith  The only invariant needed here is that the constant name starts with `XR_` so we can remove it with `len[3..]` below. I changed this so I didn't need to add the a check for `len.ends_with("_COUNT_FB") || len.starts_with("XR_UUID")`.

generator/src/main.rs
```diff
@@ -2125,18 +2135,18 @@ fn xr_var_name(raw: &str) -> Ident {
     Ident::new(&raw.to_snake_case(), Span::call_site())
 }
 
-fn xr_arg_ty(member: &Member) -> TokenStream {
+fn xr_arg_ty(api_aliases: &[(String, String)], member: &Member) -> TokenStream {
```
zjijz, line 2138: @Ralith The cost of resolving aliases was to pass the context through here and to `xr_var_ty` below. These could also become methods of `Parser` and just take `self` instead of only the `api_aliases` context.

generator/src/main.rs
zjijz: I asked FB if we could restructure the public code to use the new constant names, but that's probably a change that would take a month to get to public OpenXR-SDK.

Ralith (approved):
Thanks!

PR #108, opened 2022-02-14
Replace ndk-glue with ndk-context

zarik5:
ndk-glue provides the application entry point and was never supposed to be used by library crates. ndk-context provides the bare minimum that libraries need for plugging into the android context (VM and Context pointers); it has no dependencies and will be updated a lot less (ideally not anymore) so it becomes a stable foundation for the android crates ecosystem.

ndk-glue 0.6.1 already initializes ndk-context. This PR is not a breaking change.

### DIFF
--- a/openxr/Cargo.toml
+++ b/openxr/Cargo.toml
@@ -33,7 +33,7 @@ ctrlc = "3.1.5"
 winapi = { version = "0.3", features = ["profileapi"] }
 
 [target.'cfg(target_os = "android")'.dependencies]
-ndk-glue = "0.6"
+ndk-context = "0.1"
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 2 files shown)


kanerogers:
@zarik5 This is fantastic! `openxr` is a downstream dependency of `hotham` and any time `ndk-glue` gets bumped it causes havok. We'll switch over to `ndk-context` on `hotham` as well

zarik5:
@kanerogers Your project is really cool :) I'll keep an eye on it.
I will make a similar PR for oboe and CPAL tomorrow

kanerogers:
Fantastic! This is such a massive step forward for the Android/Rust ecosystem

Ralith (approved):
Thanks! This had been worrying me.

PR #112, opened 2022-03-28
Add support for `XR_FB_passthrough`

pierre-l:
~~*To be polished soon*~~

~~Add `create_passthrough_fb`, `create_passthrough_layer`, `new_passthrough_composition_layer`~~

Implement execution control features:
* create/destroy/pause/start passthrough,
* create/destroy/pause/resume passthrough layers

### DIFF
--- a/openxr/src/passthrough.rs
+++ b/openxr/src/passthrough.rs
@@ -1,0 +1,168 @@
+//! The [Passthrough feature](https://support.oculus.com/articles/in-vr-experiences/oculus-features/what-is-passthrough/)
+//! allows you to step outside your view in VR to see a real-time view of your surroundings.
+//!
+//! This feature is exclusive to the Oculus Quest 2 as of March 2022.
+//!
+//! More details about passthrough can be found in the [Oculus Native SDK documentation](https://developer.oculus.com/documentation/native/android/mobile-passthrough/)
+//! as well as in the [OpenXR specification](https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XrPassthroughFB)
+//!
+//! Not all features are currently implemented. Execution control features are implemented, style-oriented features are not.
+use crate::{
+    cvt, raw, session, sys, PassthroughFlagsFB, PassthroughLayerPurposeFB, Result, Session,
+    SessionInner,
+};
+use std::ptr;
+use std::sync::Arc;
+use sys::PassthroughLayerFB;
+
+/// A [passthrough feature].
+///
+/// Requires [`XR_FB_passthrough`] to be enabled.
+///
+/// See the [`PassthroughFB struct`].
+///
+/// [`XR_FB_passthrough`]: https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_FB_passthrough
+/// [`PassthroughFB struct`]: https://docs.rs/openxr-sys/latest/openxr_sys/struct.PassthroughFB.html
+/// [passthrough feature]: https://developer.oculus.com/documentation/native/android/mobile-passthrough/#create-and-start-a-passthrough-feature
+pub struct Passthrough {
+    pub(crate) session: Arc<session::SessionInner>,
+    handle: sys::PassthroughFB,
+}
+
+impl Passthrough {
+    /// [Create](https://www.khronos.org/registry/OpenXR/specs/1.0/man/html/openxr.html#xrCreatePassthroughFB)
+    /// a passthrough feature.
+    pub(crate) fn create<G>(session: &Session<G>, flags: PassthroughFlagsFB) -> Result<Self> {
+        let info = sys::PassthroughCreateInfoFB {
+            ty: sys::PassthroughCreateInfoFB::TYPE,
+            next: ptr::null(),
+            flags,
+        };
+        let fp = fp(&session.inner);
+        let mut handle = sys::PassthroughFB::NULL;
+        unsafe {
+            cvt((fp.create_passthrough)(
+                session.as_raw(),
+                &info,
+                &mut handle,
+            ))?;
+        }
+        Ok(Passthrough {
+            session: session.inner.clone(),
+            handle,
+        })
+    }
+
+    /// [Start](https://www.khronos.org/registry/OpenXR/specs/1.0/man/html/openxr.html#_xrpassthroughstartfb3)
+    /// a passthrough feature.
+    pub fn start(&self) -> Result<()> {
+        let fp = fp(&self.session);
+        unsafe {
+            cvt((fp.passthrough_pause)(self.handle))?;
+        }
+        Ok(())
+    }
+
+    /// [Pause](https://www.khronos.org/registry/OpenXR/specs/1.0/man/html/openxr.html#_xrpassthroughstartfb3)
+    /// a passthrough feature.
+    pub fn pause(&self) -> Result<()> {
+        let fp = fp(&self.session);
+        unsafe {
+            cvt((fp.passthrough_pause)(self.handle))?;
+        }
+        Ok(())
+    }
+}
+
+impl Drop for Passthrough {
+    fn drop(&mut self) {
+        unsafe {
+            (fp(&self.session).destroy_passthrough)(self.handle);
+        }
+    }
+}
+
+/// A [passthrough layer].
+///
+/// Requires [`XR_FB_passthrough`].
+///
+/// [`XR_FB_passthrough`]: https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_FB_passthrough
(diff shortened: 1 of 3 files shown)


openxr/src/passthrough.rs
```diff
@@ -0,0 +1,73 @@
...
+/// TODO Link https://developer.oculus.com/documentation/native/android/mobile-passthrough/
+/// TODO Pause/unpause
+/// Create the passthrough feature.
+///
+/// Requires [`XR_FB_passthrough`](https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_FB_passthrough)
+/// to be enabled.
+///
+/// See:
+/// * [`XR_FB_passthrough`](openxr/src/generated.rs:62)
+pub fn create_passthrough_fb<G: Graphics>(session: &Session<G>, flags: sys::PassthroughFlagsFB) -> Result<sys::PassthroughFB> {
```
Ralith, line 14: Similar extensions are exposed as methods on `Session`. We should be consistent.

openxr/src/passthrough.rs
```diff
@@ -0,0 +1,73 @@
...
+        cvt((ext.create_passthrough_layer)(session.as_raw(), &info, &mut layer))?;
+    }
+    Ok(layer)
+}
+
+/// TODO Doc
+/// TODO Link https://developer.oculus.com/documentation/native/android/mobile-passthrough/
+/// TODO Params: flags
+/// TODO Composition
+pub fn new_passthrough_composition_layer(layer: sys::PassthroughLayerFB) -> sys::CompositionLayerPassthroughFB {
```
Ralith, line 65: This should be exposed through a generated builder similar to `CompositionLayerProjection`.

openxr/src/passthrough.rs
```diff
@@ -0,0 +1,73 @@
...
+        flags
+    };
+    let ext = session
+        .instance()
+        .exts()
+        .fb_passthrough
+        .as_ref()
+        .expect("FB_passthrough not loaded");
+    let mut passthrough = sys::PassthroughFB::NULL;
+    // TODO Implement a Passthrough struct & drop?
```
Ralith, line 27: Yes.

Ralith (changes requested):

pierre-l:
@Ralith Thank you for the early feedback. I think I'm starting to understand what it should eventually look like but it's still quite raw and incomplete.

openxr/src/passthrough.rs
pierre-l: I removed this change and left the `CompositionLayerPassthroughFB` untouched. I played with the generator a bit but couldn't get satisfying results so far. Leaving this as-is doesn't prevent enabling & using passthrough, and I fear the required changes may make this PR grow a bit too much. I'll save this for later if you don't mind.

pierre-l:
@Ralith I think you'll find the updated changes much nicer and possibly merge-able.

openxr/src/passthrough.rs
```diff
@@ -0,0 +1,226 @@
...
+impl Passthrough {
+    /// [Create](https://www.khronos.org/registry/OpenXR/specs/1.0/man/html/openxr.html#xrCreatePassthroughFB) a [passthrough feature].
+    ///
+    /// # Arguments
+    ///
+    /// * `flags` - the passthrough feature flags. Unless `PassthroughFlagsFB::IS_RUNNING_AT_CREATION` is enabled, `Self::start` must be called.
+    ///
+    /// # Errors
+    ///
+    /// This function return `ERROR_EXTENSION_NOT_PRESENT` if [`XR_FB_passthrough`] is disabled.
```
Ralith, line 41: Elsewhere we just panic if the function isn't enabled, so I'd prefer to do the same here. The upstream docs also specify a wide array of other possible errors (which we probably shouldn't try to duplicate here).

Also, while docs on internal-only functions are a fine thing, we probably don't need to be as exhaustive as for public API.

openxr/src/passthrough.rs
```diff
@@ -0,0 +1,226 @@
...
+    /// [Pause](https://www.khronos.org/registry/OpenXR/specs/1.0/man/html/openxr.html#_xrpassthroughstartfb3) a [passthrough feature].
+    ///
+    /// # Errors
+    ///
+    /// This function return `ERROR_EXTENSION_NOT_PRESENT` if [`XR_FB_passthrough`] is disabled.
+    ///
+    /// [`XR_FB_passthrough`]: https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_FB_passthrough
+    /// [passthrough feature]: https://developer.oculus.com/documentation/native/android/mobile-passthrough/#create-and-start-a-passthrough-feature
+    pub(crate) fn pause(&mut self, session: &Arc<SessionInner>) -> Result<()> {
+        // Self-mutability is not obviously necessary but prevents concurrent access.
```
Ralith, line 81: Per https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#threading-behavior, OpenXR APIs are generally thread-safe and we should expose them as such.

openxr/src/passthrough.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    /// [Pause](https://www.khronos.org/registry/OpenXR/specs/1.0/man/html/openxr.html#_xrpassthroughstartfb3) a [passthrough feature].
+    ///
+    /// # Errors
+    ///
+    /// This function return `ERROR_EXTENSION_NOT_PRESENT` if [`XR_FB_passthrough`] is disabled.
+    ///
+    /// [`XR_FB_passthrough`]: https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_FB_passthrough
+    /// [passthrough feature]: https://developer.oculus.com/documentation/native/android/mobile-passthrough/#create-and-start-a-passthrough-feature
+    pub(crate) fn pause(&mut self, session: &Arc<SessionInner>) -> Result<()> {
```
Ralith, line 80: The `session` argument is redundant, as the appropriate session is already stored inside `self`.

openxr/src/session.rs
```diff
@@ -428,6 +428,77 @@ impl<G: Graphics> Session<G> {
...
+    #[inline]
+    /// Resume a [`PassthroughLayer`].
+    ///
+    /// Requires [`XR_FB_passthrough`].
+    ///
+    /// [`XR_FB_passthrough`]: https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_FB_passthrough
+    /// [`PassthroughLayer`]: passthrough/struct.PassthroughLayer.html
+    pub fn resume_passthrough_layer(&self, layer: &mut PassthroughLayer) -> Result<()> {
+        layer.resume(&self.inner)
+    }
```
Ralith, line 501: All of these save for `create_passthrough` should be methods on the appropriate new struct, not on `Session`.

Ralith (changes requested):

openxr/src/passthrough.rs
pierre-l: There's a "thread safety" mention in the `xrDestroyPassthroughFB` and `xrDestroyPassthroughLayerFB` operations spec.

```
Thread Safety

    Access to passthrough, and any child handles, must be externally synchronized
```

https://www.khronos.org/registry/OpenXR/specs/1.0/man/html/openxr.html#_xrdestroypassthroughfb3
https://www.khronos.org/registry/OpenXR/specs/1.0/man/html/openxr.html#_xrdestroypassthroughlayerfb3

This is not present in the constructors. my understanding is that the constructors themselves are thread-safe but not the returned handle.

openxr/src/passthrough.rs
pierre-l: Correct, TY!

openxr/src/passthrough.rs
pierre-l: Sometimes it does, sometimes it doesn't.
https://github.com/Ralith/openxrs/blob/master/openxr/src/foveation_profile_fb.rs#L45
https://github.com/Ralith/openxrs/blob/master/openxr/src/hand_tracker.rs#L41

I fear that if passthrough is disabled in the headset settings this call would fail. Maybe I'm wrong here. If I'm right I would like to be able to keep the program running, maybe log the error, but most importantly warn the end user with a nice dialog and a panic here would make this a lot harder.

Ralith: If passthrough is unsupported, then attempting to create an instance with it enabled must fail, which means these methods cannot be called because no instance exists. It is the application's responsibility to check for extension support and only enable those that are absolutely required or supported.

pierre-l: Sounds good. I changed this a panic and trimmed the doc a bit.

openxr/src/passthrough.rs
pierre-l: Got it. I removed the `mut`s and associated comments.

Ralith (approved):
Thanks!

PR #113, opened 2022-04-09
openxr-sys: bump cmake dependency version

philpax:
`cmake` 0.1.47 introduced VS2022 support, so I've updated the version to take advantage of that.


Ralith:
Minor and patch version updates are unnecessary, as cargo will always use the latest compatible version in new workspaces. Run `cargo update` if you need to update an old workspace.

philpax:
Hmm, fair enough - I was running into some issues where the resolution of packages led to a pair of (`cmake`, `cc`) versions being selected that would result in my local VS2022 being selected without support from the cmake package when building openxrs. I can't seem to reproduce that now, but hopefully it was just a fleeting issue :pray:

PR #114, opened 2022-04-14
Add support for `XR_FB_display_refresh_rate`

Timbals:


openxr/src/display_refresh_rate.rs
```diff
@@ -0,0 +1,67 @@
...
+        let ext = self
+            .inner
+            .instance
+            .exts()
+            .fb_display_refresh_rate
+            .as_ref()
+            .expect("XR_FB_display_refresh_rate not loaded");
+        unsafe {
+            let mut out = MaybeUninit::uninit();
+            (ext.get_display_refresh_rate)(self.as_raw(), out.as_mut_ptr());
```
Ralith, line 46:
```suggestion
            cvt((ext.get_display_refresh_rate)(self.as_raw(), out.as_mut_ptr()))?;
```

openxr/src/display_refresh_rate.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub fn request_display_refresh_rate(&self, display_refresh_rate: f32) -> Result<()> {
+        let ext = self
+            .inner
+            .instance
+            .exts()
+            .fb_display_refresh_rate
+            .as_ref()
+            .expect("XR_FB_display_refresh_rate not loaded");
+        cvt(unsafe { (ext.request_display_refresh_rate)(self.as_raw(), display_refresh_rate) })?;
+        Ok(())
```
Ralith, line 65: nit:
```suggestion
        cvt(unsafe { (ext.request_display_refresh_rate)(self.as_raw(), display_refresh_rate) })
```

Ralith (changes requested):
Thanks!

openxr/src/display_refresh_rate.rs
Timbals: This doesn't work because of the return type being `Result<()>` and not `Result<sys::Result>`.

Ralith (approved):

PR #115, opened 2022-04-25
Remove `khr_vulkan_enable` support in example

zarik5:
All major XR runtimes have upgraded to `khr_vulkan_enable2` and new runtimes implement `khr_vulkan_enable2` directly. This change makes the example more concise and straightforward. `khr_vulkan_enable` support is kept in `openxrs` API for legacy reasons.

### DIFF
--- a/openxr/examples/vulkan.rs
+++ b/openxr/examples/vulkan.rs
@@ -6,7 +6,6 @@
 //! largely decouple its Vulkan and OpenXR components and handle errors gracefully.
 
 use std::{
-    ffi::{CStr, CString},
     io::Cursor,
     sync::{
         atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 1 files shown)


Ralith (approved):
Thanks!

kanerogers:
Is this correct? The Oculus Quest (the first one) did not support enable2
last time I checked.

Ralith:
It'd be interesting to have confirmation, but if someone really needs to target that, they can always use an old release or the raw API. This being a volunteer project I have limited bandwidth to support hardware that even first-party support has been dropped for.

kanerogers:
I've had a think about this, @Ralith and I think you're right. 👍🏻 

We'll move to `khr_vulkan_enable2` in Hotham and drop support for the Quest.

PR #120, opened 2022-08-09
Add features for layer development

Sorenon:
This makes two minor changes to make openxrs more practical for developing an OpenXR layer.

Adding Hash to raw handle types makes it simpler to manage all the tracked handles like so:
```rs
static mut INSTANCE_WRAPPERS: OnceCell<DashMap<xr::Instance, Arc<instance::InstanceWrapper>>> = OnceCell::new();
```

And the added bindings from [loader_interfaces.h](https://github.com/KhronosGroup/OpenXR-SDK-Source/blob/main/src/common/loader_interfaces.h) are used for Runtime-Loader-Layer negotiation.

### DIFF
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -1031,7 +1031,7 @@ impl Parser {
             quote! {
                 #[doc = #doc]
                 #[repr(transparent)]
-                #[derive(Debug, Copy, Clone, Eq, PartialEq)]
+                #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
                 pub struct #ident(u64);
                 handle!(#ident);
             }
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -1191,7 +1191,7 @@ impl Parser {
         quote! {
             //! Automatically generated code; do not edit!
 
-            #![allow(non_upper_case_globals, clippy::unreadable_literal, clippy::identity_op, unused)]
+            #![allow(non_upper_case_globals, clippy::unreadable_literal, clippy::identity_op, unused, clippy::derive_partial_eq_without_eq)]
             use std::fmt;
             use std::mem::MaybeUninit;
             use std::os::raw::{c_void, c_char};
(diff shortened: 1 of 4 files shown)


sys/src/loader_interfaces.rs
```diff
@@ -0,0 +1,153 @@
...
+    pub const VERSION: u32 = 1;
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct XrNegotiateRuntimeRequest {
+    pub ty: LoaderInterfaceStructureType,
+    pub struct_version: u32,
+    pub struct_size: usize,
+    /// CURRENT_LOADER_RUNTIME_VERSION
```
Ralith, line 88: What's this comment meant to communicate?

Ralith (commented):
Thanks! Would it make sense to generate these, rather than hand-writing them?

sys/src/loader_interfaces.rs
Sorenon: Its more or less directly copied from the header file, it's meant to communicate that the runtime should set `runtime_interface_version` to `CURRENT_LOADER_RUNTIME_VERSION`.
I overlooked it because it's not directly involved in layer negotiation but if you want I can go over the file again and make the docs more self explanatory.

Sorenon:
> Thanks! Would it make sense to generate these, rather than hand-writing them?

Since the file is so small and not defined by XML I think that writing a generator would be overkill.

Ralith:
Ah, yeah, if there's no XML source then the effort isn't justified. Nice job keeping the style consistent!

sys/src/loader_interfaces.rs
Ralith: Let's either omit it or make it useful; your call. Generally I'm fine with pointing people at upstream docs, and that goes double for niche advanced use cases like layer authoring.

Sorenon: Got it. I'll just remove it for now with the possibility of of adding more detailed documentation in future.

Ralith:
Sorry this languished; Actions bugged out somehow.

PR #122, opened 2022-09-24
fix(generator): use manifest dir for output files

philpax:
extremely minor fix for an issue I encountered while trying to edit the generator for something else - `cargo run -p generator` now works as expected at the root


Ralith (approved):
Thanks!

PR #124, opened 2022-10-02
added wgpu example

Gonkalbell:
This is mostly an experiment on my part to see if I could get WGPU to work on the swapchain provided by OpenXR, using WGPU-HAL's functions for working with raw Vulkan objects. I think this may be useful enough to merge into the repo in case other developers are trying to do the same thing. This example only works if you use Vulkan as a backend, but it can probably be updated later to use DX12 and Metal as well. I did have to stumble through some of the sparsely documented parts of WGPU, so I could definitely use some feedback to make sure I dot all my "i"s and cross all my "t"s.

### DIFF
--- a/openxr/examples/wgpu.rs
+++ b/openxr/examples/wgpu.rs
@@ -1,0 +1,647 @@
+use std::{
+    num::NonZeroU32,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
+
+use ash::{
+    vk::{self, Handle},
+};
+use openxr as xr;
+
+use wgpu_hal::{api::Vulkan as HalVulkan, Api};
+
+pub const COLOR_FORMAT: wgpu::TextureFormat = wgpu::TextureFormat::Rgba8UnormSrgb;
+pub const VIEW_COUNT: u32 = 2;
+const VIEW_TYPE: xr::ViewConfigurationType = xr::ViewConfigurationType::PRIMARY_STEREO;
+
+/// Maximum number of frames in flight
+const PIPELINE_DEPTH: u32 = 2;
+
+struct Swapchain {
+    handle: xr::Swapchain<xr::Vulkan>,
+    width: u32,
+    height: u32,
+    textures: Vec<wgpu::Texture>,
+}
+
+fn main() {
+    // Handle interrupts gracefully
+    let running = Arc::new(AtomicBool::new(true));
+    let r = running.clone();
+    ctrlc::set_handler(move || {
+        r.store(false, Ordering::Relaxed);
+    })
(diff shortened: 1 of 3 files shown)


openxr/examples/wgpu.rs
```diff
@@ -0,0 +1,645 @@
...
+            .application_version(0)
+            .engine_version(0)
+            .api_version(vk_target_version);
+
+        let vk_instance = {
+            let vk_instance = xr_instance
+                .create_vulkan_instance(
+                    system,
+                    std::mem::transmute(vk_entry.static_fn().get_instance_proc_addr),
+                    &vk::InstanceCreateInfo::builder().application_info(&vk_app_info) as *const _
```
Noxime, line 130: You should account for `HalVulkan::required_extensions` when creationg the Vulkan instance (enabled_extension_names)

openxr/examples/wgpu.rs
```diff
@@ -0,0 +1,645 @@
...
+            .expect("Vulkan device has no graphics queue");
+
+        let vk_device = {
+            let vk_device = xr_instance
+                .create_vulkan_device(
+                    system,
+                    std::mem::transmute(vk_entry.static_fn().get_instance_proc_addr),
+                    vk_physical_device.as_raw() as _,
+                    &vk::DeviceCreateInfo::builder()
+                        .queue_create_infos(&[vk::DeviceQueueCreateInfo::builder()
```
Noxime, line 174: `enabled_extension_names` should be `ExposedAdapter::adapter::required_device_extensions`

Noxime (changes requested):
This is very useful, I know a few people are interested in using WGPU with OpenXR. Just a few notes about making sure the correct instance and device extensions are loaded.  [Heres a gist](https://gist.github.com/Noxime/ba570ca38c4e91470ddd34a3e23babd4) of my code to initialize WGPU from OpenXR

zarik5:
I just want to mention I did a similar thing a while ago, but I took some shortcuts (no multiview and i reused the same swapchain for both eyes which does not make sense), and the wgpu version is old: https://github.com/zarik5/openxrs/tree/wgpu-test
Maybe it can be useful.

openxr/examples/wgpu.rs
Gonkalbell: I just updated this to check for extensions on the device and instance. I was originally copying the Vulkan example, which didn't seam to check those. But I think your right and I should be checking them

LoganDark:
image

WMR OpenXR does not support Vulkan at all, is there any way to use DX11 or DX12? I know wgpu supports DirectX as well...

LoganDark:
Alright, soooooo,,, this crashes with an access violation if I go to the WMR Home. No idea how or why.

PR #125, opened 2022-10-04
update OpenXR-SDK to 1.0.25

kcking:
fixes macOS build with https://github.com/KhronosGroup/OpenXR-SDK-Source/pull/323


Ralith:
Thanks! Want to run the generator for the update too?

kcking:
Ah yes that was too easy... 😅 forgot about the generator.

@Ralith I added two commits to fix the build errors, except that `VkImageCreateFlags` and `VkImageUsageFlags` are undefined. It appears OpenXR-SDK assumes these are imported from vulkan.h. `ash` is currently a dependency of `openxr` but not `sys`. Should I add `ash` as a dep of `sys` and either 1) typedef or 2) substitute those types for the associated `ash::vk::` types?

Ralith:
No, `ash` should not be a dependency of either crate. The new Vulkan types should be handled consistently with the existing ones, by transcribing the bare minimum representation into `sys/src/platform.rs`. This ensures we don't break every time `ash` updates.

Generating atom definitions automatically is nice, but should probably be pursued in a separate PR. Note that we currently have handwritten definitions for most of them in `sys/src/lib.rs`.

generator/src/main.rs
```diff
@@ -2126,6 +2126,9 @@ fn xr_bitmask_value_name(ty: &str, name: &str) -> Ident {
     let ty = &ty[0..ty.len() - "Flags".len()];
     let prefix_len = ty.to_shouty_snake_case().len() + 1;
     let end = name.rfind("_BIT").unwrap();
+    if prefix_len == end + 1 {
+        return Ident::new(name, Span::call_site());
+    }
```
Ralith, line 2184: For consistency and concision, we should generate e.g. `PASSTHROUGH_CAPABILITY` here, omitting the redundant `XR_` and `_BIT` bookends. A comment justifying the logic would also be nice.

kcking:
Great, thanks for the comments! Latest version now pushed.
I have it down to some new FB spatial anchor/Space struct issues.
1. they use a `UuidEXT` struct that holds a 32-byte array. should we create a wrapper for this, or just export it as-is? If users wanted to manipulate the UUIDs instead of just using them as opaque handles, passing them to the `uuid` crate seems easy enough. `UuidEXT` currently fails the `&& x.static_array_len.is_none()` check of `is_simple_struct`.
2. some generated struct wrappers like `SpaceQueryCompleteFB` turn `XrResult` into `Result` which is then conflicting with `std::Result`. do we have an existing solution for this? should we be mapping `XrResult` to `Result<()>`?

Thanks for your help!

kcking:
Applied work-arounds for both of the above issues, let me what you think :)

generator/src/main.rs
```diff
@@ -1379,6 +1379,8 @@ impl Parser {
                 }
             })
             .collect::<IndexSet<&str>>();
+
+        let forced_reexports = ["UuidEXT"];
```
Ralith, line 1383: Rather than hardcoding static stuff into the generator, it's easier and more maintainable to add it to the appropriate lib.rs directly.

Ralith (changes requested):

openxr/src/generated.rs
```diff
@@ -2365,6 +2794,12 @@ pub mod raw {
         pub const NAME: &'static [u8] = sys::EXT_HP_MIXED_REALITY_CONTROLLER_EXTENSION_NAME;
     }
     #[derive(Copy, Clone)]
+    pub struct PalmPoseEXT {}
+    impl PalmPoseEXT {
+        pub const VERSION: u32 = sys::EXT_palm_pose_SPEC_VERSION;
+        pub const NAME: &'static [u8] = sys::EXT_PALM_POSE_EXTENSION_NAME;
+    }
+    #[derive(Copy, Clone)]
     pub struct UuidEXT {}
```
Ralith, line 2803: It looks like this isn't actually being generated correctly to begin with, unfortunately.

kcking: so the extension and the actual UUID struct have the same name. the uuid struct is generated [here](https://github.com/kcking/openxrs/blob/a385cf4c2329a002e95d943b41e6dcdd2b49f2c2/sys/src/generated.rs#L6849) which looks right to me

Ralith: NVM, this is the extension struct, lol.

Ralith (approved):
Thanks!

Ralith:
Published in v0.17.1.

PR #131, opened 2023-01-29
Derive Copy and Clone for Graphics structs

Supreeeme:
Allows types that use these to also be cloned and/or copied.

### DIFF
--- a/openxr/src/graphics/d3d.rs
+++ b/openxr/src/graphics/d3d.rs
@@ -9,6 +9,7 @@ use crate::*;
 /// See [`XR_KHR_d3d11_enable`] for safety details.
 ///
 /// [`XR_KHR_d3d_enable`]: https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_KHR_D3D11_enable
+#[derive(Copy, Clone)]
 pub enum D3D11 {}
 
 impl Graphics for D3D11 {
--- a/openxr/src/graphics/opengl.rs
+++ b/openxr/src/graphics/opengl.rs
@@ -9,6 +9,7 @@ use crate::*;
 /// See [`XR_KHR_opengl_enable`] for safety details.
 ///
 /// [`XR_KHR_opengl_enable`]: https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_KHR_opengl_enable
+#[derive(Copy, Clone)]
 pub enum OpenGL {}
 
 impl Graphics for OpenGL {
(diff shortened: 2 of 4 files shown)


Ralith:
This isn't strictly needed, since these types, having no values, can never be constructed to begin with. I'm also not sure how this could affect any generic code regardless, since there's no change to the trait definition.

Supreeeme:
example:
```rust
use openxr as xr;
fn main() {
    let layer = xr::CompositionLayerQuad::<xr::Vulkan>::new();
    let layer2 = layer.clone();
}
```
Without these changes I get these errors:
```rust
error[E0599]: the method `clone` exists for struct `openxr::CompositionLayerQuad<'_, Vulkan>`, but its trait bounds were not satisfied
    --> src/main.rs:4:24
     |
4    |     let layer2 = layer.clone();
     |                        ^^^^^ method cannot be called on `openxr::CompositionLayerQuad<'_, Vulkan>` due to unsatisfied trait bounds
     |
    ::: /home/supreme/.cargo/registry/src/github.com-1ecc6299db9ec823/openxr-0.17.1/src/generated.rs:4710:5
     |
4710 |     pub struct CompositionLayerQuad<'a, G: Graphics> {
     |     ------------------------------------------------ doesn't satisfy `openxr::CompositionLayerQuad<'_, Vulkan>: Clone`
     |
    ::: /home/supreme/.cargo/registry/src/github.com-1ecc6299db9ec823/openxr-0.17.1/src/graphics/vulkan.rs:12:1
     |
12   | pub enum Vulkan {}
     | --------------- doesn't satisfy `Vulkan: Clone`
     |
     = note: the following trait bounds were not satisfied:
             `Vulkan: Clone`
             which is required by `openxr::CompositionLayerQuad<'_, Vulkan>: Clone`
```

Ralith:
So, the real bug here is that `CompositionLayerQuad` (and similar structs) should not be using `#[derive(Clone)]`, but rather a manual impl that doesn't constrain the type variable. A fix for that would be welcome.

That said, it's a bit strange to be constructing freestanding composition layers like that, and to be cloning them. They're only meant to serve as a container for arguments passed to `FrameStream::end`, as in the associated example code.

Supreeeme:
I'm constructing two similar quad layers (only differing in swapchain) so it made sense to just clone one. I'll see if I can get around to fixing the clone impl properly.

Ralith:
I'd also be receptive to refactoring to make graphics structs ZSTs that get stored directly rather than using `PhantomData`; that would allow derives to work and might involve less boilerplate in the end, as well as dovetailing gracefully with some experimental wgpu compatibility work others have been playing with.

PR #138, opened 2023-07-13
early Android OpenGLES support, tested on Quest 2

mutantbob:
This is a `Graphics` implementation that seems to work on my Quest 2 VR headset.

I have used it with the project at https://github.com/mutantbob/android-openxr-exp and it seems to work.

It is time to get it in front of more developers so it can be tuned and tested on more headsets.


Ralith:
How does this differ from the existing `OpenGlEs` integration?

zarik5:
I'm using the existing OpenGL implementation in openxrs without issues on the Quest 2 (and other android headsets). Here is my project: https://github.com/alvr-org/ALVR/blob/master/alvr/client_openxr/src/lib.rs

PR #139, opened 2023-09-12
Update OpenXR-SDK to 1.0.29

rasmusgo:
This is what happened when I wanted to be able to use `XR_FB_haptic_pcm` and went down the rabbit hole. There was a series of issues that I bumped into. Some are described in #134 but I chose to omit extensions with unlisted tags instead of allowing them (`XR_EXTX_overlay` etc).

On a related note, this is the most hacky code I have written in a while. I am grateful for any hints on how to improve it.

### DIFF
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -600,7 +600,7 @@ impl Parser {
             }
             self.base_headers
                 .entry(parent.into())
-                .or_insert_with(Vec::new)
+                .or_default()
                 .push(struct_name.into());
         }
         if let Some(target) = attr(attrs, "alias") {
@@ -1135,7 +1135,7 @@ impl Parser {
         let commands = self.commands.iter().chain(
             self.cmd_aliases
                 .iter()
-                .map(|&(ref name, ref target)| (name, self.commands.get(target).unwrap())),
+                .map(|(name, target)| (name, self.commands.get(target).unwrap())),
         );
 
         let (pfns, protos) = commands
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -78,6 +78,7 @@ impl Parser {
             extensions: IndexMap::new(),
             // TODO: Handle these extensions
             disabled_exts: [
+                "XR_HTC_passthrough",
                 "XR_MSFT_scene_understanding",
                 "XR_MSFT_scene_understanding_serialization",
             ]
@@ -312,16 +313,16 @@ impl Parser {
         if attr(attrs, "supported").map_or(false, |x| x == "disabled") {
             self.disabled_exts.insert(ext_name);
         } else {
-            let (tag, _) = split_ext_tag(&ext_name);
-            self.extensions
-                .get_mut(tag)
-                .unwrap()
-                .extensions
-                .push(Extension {
+            let (tag_name, _) = split_ext_tag(&ext_name);
+            if let Some(tag) = self.extensions.get_mut(tag_name) {
+                tag.extensions.push(Extension {
                     name: ext_name,
                     version: ext_version.unwrap(),
                     commands,
                 });
+            } else {
+                eprintln!("ignoring extension with unlisted tag: {}", ext_name);
+            }
         }
     }
 
@@ -546,6 +547,11 @@ impl Parser {
     }
 
     fn parse_struct(&mut self, attrs: &[OwnedAttribute]) {
+        const STRUCTS_BLACKLIST: [&str; 3] = [
+            "XrCompositionLayerPassthroughHTC",
+            "XrPassthroughColorHTC",
+            "XrPassthroughHTC",
+        ];
         let struct_name = attr(attrs, "name").unwrap();
         let mut members = Vec::new();
         let mut ty = None;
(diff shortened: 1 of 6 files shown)


generator/src/main.rs
```diff
@@ -546,6 +547,11 @@ impl Parser {
     }
 
     fn parse_struct(&mut self, attrs: &[OwnedAttribute]) {
+        const STRUCTS_BLACKLIST: [&str; 3] = [
+            "XrCompositionLayerPassthroughHTC",
+            "XrPassthroughColorHTC",
+            "XrPassthroughHTC",
+        ];
```
rasmusgo, line 554: This might be the worst part.

Ralith: Is there a way we could rely on `disabled_exts` exclusively to mask these?

generator/src/main.rs
```diff
@@ -1601,10 +1621,14 @@ impl Parser {
             out.has_pointer |= member.ptr_depth != 0 || self.handles.contains(&member.ty);
             out.has_graphics |= member.ty == "XrSession" || member.ty == "XrSwapchain";
             out.has_array |= member.static_array_len.is_some();
+            out.has_non_default |= member.ty == "XrTime" || member.ty == "XrDuration";
```
Ralith, line 1624: Should we impl `Default` for these instead? At least for `Duration` it seems totally reasonable. Maybe not `Time`, per std's precedent on `Instant`.

Ralith (commented):
LGTM overall, couple possible improvements

generator/src/main.rs
rasmusgo: I added `XR_HTC_passthrough` to `disabled_exts` but it wasn't sufficient. The xml structure is a bit backwards in this regard as the types are defined before the extension is named and the extension is only referring to these by name. We would need to parse the extensions before the structs and dynamically build a blacklist or remove the structs after parsing the extensions.

openxr/src/generated.rs
```diff
@@ -5216,6 +6395,140 @@ pub(crate) mod builder {
...
+            self.inner.sample_rate = value;
+            self
+        }
+        #[inline]
+        pub fn append(mut self, value: bool) -> Self {
+            self.inner.append = value.into();
+            self
+        }
+        #[inline]
+        pub fn samples_consumed(mut self, value: &'a u32) -> Self {
```
rasmusgo, line 6515: This value is modified by the runtime to report back the number of samples that were consumed. It would be great if this could be reflected by the type somehow. Are there any other examples of that?

generator/src/main.rs
Ralith: Is it the parsing that fails, or the generator? If the latter, then removing structs after parsing extensions seems like a somewhat more maintainable solution.

openxr/src/generated.rs
Ralith: Good catch; in fact we *must* arrange for this to be `&mut`, or use will trigger UB. Can we select `&` vs `&mut` based on whether the pointee is `const`?

generator/src/main.rs
rasmusgo: The generator completes but the build fails afterwards so we should be able to prune the list of structs after parsing is done.

openxr/src/generated.rs
rasmusgo: I think that should work. Will try!

rasmusgo: Looks like it's working!

openxr/src/lib.rs
```diff
@@ -44,7 +44,7 @@ pub use passthrough::*;
 pub use builder::{
     CompositionLayerBase, CompositionLayerCubeKHR, CompositionLayerCylinderKHR,
     CompositionLayerEquirectKHR, CompositionLayerProjection, CompositionLayerProjectionView,
-    CompositionLayerQuad, HapticBase, HapticVibration, SwapchainSubImage,
+    CompositionLayerQuad, HapticBase, HapticPcmVibrationFB, HapticVibration, SwapchainSubImage,
```
rasmusgo, line 47: These are easy to forget. Can we use `builder::*` instead?

openxr/src/generated.rs
Ralith: Awesome! I guess this really was the first occurrence of an output parameter.

openxr/src/lib.rs
rasmusgo: Here is an incomplete list of things we are not exporting from `builder`:

Ralith: IIRC this is currently intended to be an allowlist: we only export builders that the high-level API consumes.

generator/src/main.rs
Ralith: That makes sense, and sounds like an ideal balance. Users being able to fall back on the unsafe raw bindings is a key design goal.

generator/src/main.rs
```diff
@@ -1621,7 +1621,7 @@ impl Parser {
             out.has_pointer |= member.ptr_depth != 0 || self.handles.contains(&member.ty);
             out.has_graphics |= member.ty == "XrSession" || member.ty == "XrSwapchain";
             out.has_array |= member.static_array_len.is_some();
-            out.has_non_default |= member.ty == "XrTime" || member.ty == "XrDuration";
```
Ralith: Did you forget to add the `impl Default` in this commit?

rasmusgo: It was actually there already

rasmusgo: Sorry, I must have mixed it up with something else. I have added a derive for it now.

PR #143, opened 2023-11-16
Update OpenXR-SDK to 1.0.31

rasmusgo:
No big changes since 1.0.29 but nice to keep up to date.

I'm eagerly waiting for the Quest 3 upper body tracking extension that is expected to come in December so I am being a bit proactive to reduce the risk of blockers.


Ralith (approved):
Thanks! Good to not let this stuff back up too much.

PR #145, opened 2023-11-21
Added links field to cargo.toml

TheButlah:
Generally, -sys crates should specify the [links field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-links-field). This is important because cargo provides mechanisms to [override the linker search path](https://doc.rust-lang.org/cargo/reference/config.html#targettriplelinks) using this field.

[Example usage](https://github.com/NexusSocial/nexus-vr/pull/3)


Ralith (approved):
Thanks!

PR #146, opened 2023-12-01
Add basic D3D12 bindings

SafariMonkey:
This PR copies the D3D11 bindings with minor modifications to work for D3D12. It at least works for me as used in awtterpip/bevy_oxr/pull/50, but there may be some subtleties that I've missed.


Ralith (approved):
LGTM, thanks!

PR #147, opened 2023-12-08
Update to OpenXR 1.0.32.1

rasmusgo:


openxr/src/generated.rs
```diff
@@ -2960,6 +3029,38 @@ impl<'a> EyeCalibrationChangedML<'a> {
...
+    #[inline]
+    pub fn session(self) -> sys::Session {
+        (self.0).session
+    }
+    #[inline]
+    pub fn state(self) -> LocalizationMapStateML {
+        (self.0).state
+    }
+    #[inline]
+    pub fn map(self) -> LocalizationMapML {
```
rasmusgo, line 3052: This is a bit questionable because it contains a name field with a null terminated string. Should we put some effort into transforming it to a regular rust string?

Ralith: I think that'd be ideal. That might be why we don't generate the reexport by default.

rasmusgo: I made an attempt to fix it now. Please scrutinize appropriately!

openxr/src/localization_map_ml.rs
Ralith, line 1: Most high level structs use public fields rather than getters. Can we be consistent with that here?

Ralith:
@rasmusgo friendly ping?

rasmusgo:
Sorry for the delay, I pushed a commit for the public fields instead of getters now.

rasmusgo:
@Ralith is this PR still relevant?

PR #151, opened 2024-02-02
Monado EGL extension

technobaboo:
so I need help to get the extension into the list so it can be enabled but i got the graphics bindings structs in


technobaboo:
@Ralith idk how the generator works, got any ideas? this is needed for servo webxr too

Ralith:
Is it sufficient for applications to manually add it to the `other` list for the time being?

rasmusgo:
It looks like this was fixed by https://github.com/Ralith/openxrs/pull/156. @technobaboo did it resolve it for you?

PR #157, opened 2024-04-23
Update to OpenXR 1.0.34

rasmusgo:
Here are some relatively simple changes before going all the way to OpenXR 1.1.36.

### DIFF
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -1261,8 +1261,13 @@ impl Parser {
     /// Generate high-level code
     #[allow(clippy::cognitive_complexity)] // TODO
     fn generate_hl(&self) -> TokenStream {
+        const BLACKLISTED_COMMANDS: [&str; 3] = [
+            "xrNegotiateLoaderRuntimeInterface",
+            "xrNegotiateLoaderApiLayerInterface",
+            "xrCreateApiLayerInstance",
+        ];
         let (instance_pfn_fields, instance_pfn_inits) = self.commands.iter().map(|(name, command)| {
-            if command.extension.is_some() {
+            if command.extension.is_some() || BLACKLISTED_COMMANDS.contains(&name.as_str()) {
                 return (quote! {}, quote! {});
             }
             let pfn_ident = xr_command_name(name);
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -1115,6 +1115,14 @@ impl Parser {
             } else {
                 quote! { #[derive(Copy, Clone, Debug, Default, PartialEq)] }
             };
+            let aliases = self.struct_aliases.iter().map(|(alias, alias_to)| {
+                if alias_to == name {
+                    let alias_ident = xr_ty_name(alias);
+                    quote! { pub type #alias_ident = #ident; }
+                } else {
+                    quote! {}
+                }
+            });
             quote! {
                 #[repr(C)]
                 #derives
@@ -1124,6 +1132,7 @@ impl Parser {
                     #(#members)*
                 }
                 #ty
+                #(#aliases)*
             }
         });
 
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -1116,12 +1116,11 @@ impl Parser {
                 quote! { #[derive(Copy, Clone, Debug, Default, PartialEq)] }
             };
             let aliases = self.struct_aliases.iter().map(|(alias, alias_to)| {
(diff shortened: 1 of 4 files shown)


openxr/src/generated.rs
```diff
@@ -3127,6 +3129,9 @@ pub mod raw {
     use sys::pfn;
     #[derive(Copy, Clone)]
     pub struct Instance {
+        pub negotiate_loader_runtime_interface: pfn::NegotiateLoaderRuntimeInterface,
+        pub negotiate_loader_api_layer_interface: pfn::NegotiateLoaderApiLayerInterface,
+        pub create_api_layer_instance: pfn::CreateApiLayerInstance,
```
Ralith, line 3250: I don't think these are supposed to be application-facing. Should we be exposing them?

generator/src/main.rs
```diff
@@ -1115,6 +1115,14 @@ impl Parser {
             } else {
                 quote! { #[derive(Copy, Clone, Debug, Default, PartialEq)] }
             };
+            let aliases = self.struct_aliases.iter().map(|(alias, alias_to)| {
+                if alias_to == name {
```
Ralith, line 1119: style nit: invert condition and return early to reduce rightward drift

Ralith (changes requested):

generator/src/main.rs
rasmusgo: How would that work with `quote! {}` here?

openxr/src/generated.rs
rasmusgo: I'm not sure how they compare to the rest of the function pointers. What are the trade-offs?

generator/src/main.rs
Ralith:
```rs
if alias_to != name {
    return quote! {};
}
quote! { ... }
```

openxr/src/generated.rs
Ralith: Other function pointers are intended to be exposed by the loader and called by applications. These are, at a glance, intended to be exposed by runtimes/layers and called by the loader.

rasmusgo:
- [ ] Links to `https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html` should be updated to 1.1.

openxr/src/generated.rs
rasmusgo: Ok, I added a blacklist to remove these from the high level bindings.

Ralith: Is there nothing at all distinguishing them in the XML?

rasmusgo: Only a comment above that block of commands.

rasmusgo:
```xml
    <!-- SECTION: OpenXR loader command definitions -->
    <!-- openxr_loader_negotiation.h -->
    <commands>
        <command successcodes="XR_SUCCESS" errorcodes="XR_ERROR_INITIALIZATION_FAILED">
            <proto><type>XrResult</type> <name>xrNegotiateLoaderRuntimeInterface</name></proto>
            <param>const <type>XrNegotiateLoaderInfo</type>* <name>loaderInfo</name></param>
            <param><type>XrNegotiateRuntimeRequest</type>* <name>runtimeRequest</name></param>
        </command>
        <command successcodes="XR_SUCCESS" errorcodes="XR_ERROR_INITIALIZATION_FAILED">
            <proto><type>XrResult</type> <name>xrNegotiateLoaderApiLayerInterface</name></proto>
            <param>const <type>XrNegotiateLoaderInfo</type>* <name>loaderInfo</name></param>
            <param len="null-terminated">const <type>char</type>* <name>layerName</name></param>
            <param><type>XrNegotiateApiLayerRequest</type>* <name>apiLayerRequest</name></param>
        </command>
        <command successcodes="XR_SUCCESS" errorcodes="XR_ERROR_INITIALIZATION_FAILED">
            <proto><type>XrResult</type> <name>xrCreateApiLayerInstance</name></proto>
            <param>const <type>XrInstanceCreateInfo</type>* <name>info</name></param>
            <param>const <type>XrApiLayerCreateInfo</type>* <name>layerInfo</name></param>
            <param><type>XrInstance</type>* <name>instance</name></param>
        </command>
    </commands>
```

Ralith: That's annoying. Guess we'll just have to stay vigilant.

PR #158, opened 2024-04-23
Update to OpenXR 1.1.36

rasmusgo:
This PR builds upon #157.

The biggest change in the generator is that `<feature>` tags are processed now. They are similar to `<extension>` but have a couple of important differences.

`<interaction_profiles>` were added in OpenXR 1.0.27 but we are still not using them. OpenXR 1.1.36 keeps building upon them with `<extend>` and `<interaction_profile>` tags inside `<require>` tags. I think we should consider adding support for them in another PR.

I used https://registry.khronos.org/OpenXR/specs/1.1/styleguide.html#schema and https://registry.khronos.org/vulkan/specs/1.3/registry.html to figure out the schema.

### DIFF
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -110,6 +110,9 @@ impl Parser {
                     "commands" => {
                         self.parse_commands();
                     }
+                    "feature" => {
+                        self.parse_feature();
+                    }
                     "extensions" => {
                         self.parse_extensions();
                     }
@@ -150,6 +153,33 @@ impl Parser {
                 },
                 EndElement { name } => {
                     if name.local_name == "tags" {
+                        break;
+                    }
+                    eprintln!("unexpected end element: {}", name);
+                }
+                EndDocument => {
+                    panic!("unexpected end of document");
+                }
+                _ => {}
+            }
+        }
+    }
+
+    fn parse_feature(&mut self) {
+        loop {
+            use XmlEvent::*;
+            match self.reader.next().expect("failed to parse XML") {
+                StartElement { name, .. } => match &name.local_name[..] {
+                    "require" => {
+                        self.parse_required(None, None);
+                    }
+                    _ => {
+                        eprintln!("unimplemented feature element: {}", name.local_name);
+                        self.finish_element();
+                    }
+                },
+                EndElement { name } => {
+                    if name.local_name == "feature" {
                         break;
                     }
                     eprintln!("unexpected end element: {}", name);
@@ -205,7 +235,7 @@ impl Parser {
     fn parse_extension_required(&mut self, attrs: &[OwnedAttribute]) {
         let ext_name = Rc::<str>::from(attr(attrs, "name").unwrap());
         let ext_number = attr(attrs, "number").unwrap().parse::<i32>().unwrap();
-        let (ext_version, commands) = self.parse_required(&ext_name, ext_number);
+        let (ext_version, commands) = self.parse_required(Some(&ext_name), Some(ext_number));
         if attr(attrs, "supported").map_or(false, |x| x == "disabled") {
             self.disabled_exts.insert(ext_name);
         } else {
@@ -235,9 +265,10 @@ impl Parser {
 
     fn parse_required(
         &mut self,
-        ext_name: &Rc<str>,
-        ext_number: i32,
+        ext_name: Option<&Rc<str>>,
+        ext_number: Option<i32>,
     ) -> (Option<u32>, Vec<String>) {
+        assert_eq!(ext_name.is_some(), ext_number.is_some());
         let mut ext_version = None;
         let mut commands = Vec::new();
         loop {
@@ -248,8 +279,10 @@ impl Parser {
                 } => match &name.local_name[..] {
                     "command" => {
                         let cmd = attr(&attributes, "name").unwrap();
-                        if let Some(command) = self.commands.get_mut(cmd) {
-                            command.extension = Some(ext_name.clone());
+                        if let Some(ext_name) = ext_name {
+                            if let Some(command) = self.commands.get_mut(cmd) {
+                                command.extension = Some(ext_name.clone());
+                            }
                         }
                         commands.push(cmd.into());
                         self.finish_element();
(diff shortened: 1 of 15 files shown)


generator/src/main.rs
```diff
@@ -162,6 +165,125 @@ impl Parser {
...
+                }
+                EndDocument => {
+                    panic!("unexpected end of document");
+                }
+                _ => {}
+            }
+        }
+    }
+
+    fn parse_feature_required(&mut self) {
```
Ralith, line 195: This looks like it has a lot of copy-paste from `parse_extension_required`. Can we factor that out?

Ralith (changes requested):

generator/src/main.rs
rasmusgo: I was able to unify them into `parse_required` by adding parameters to optionally pass extension name and number.

rasmusgo:
- [ ] Links to `https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html` should be updated to 1.1.

awtterpip:
is there any ETA on this getting merged?

Ralith:
It needs a rebase first, for one thing.

rasmusgo:
Rebased now. The most glaring problem is that the instance cannot be created if OpenXR 1.1 isn't supported by the loader, as we try to use the current API version when creating the instance: https://github.com/Ralith/openxrs/blob/4681a25c910da1f078342935b60b4fabe3cc4747/openxr/src/entry.rs#L239

Ralith:
Would it make sense to take a version argument, deferring that problem to the application? That'd help us gracefully handle movement of extensions that we provide high-level wrappers for into core, too.

generator/src/main.rs
```diff
@@ -1308,13 +1317,10 @@ impl Parser {
     /// Generate high-level code
     #[allow(clippy::cognitive_complexity)] // TODO
     fn generate_hl(&self) -> TokenStream {
-        const BLACKLISTED_COMMANDS: [&str; 3] = [
-            "xrNegotiateLoaderRuntimeInterface",
-            "xrNegotiateLoaderApiLayerInterface",
-            "xrCreateApiLayerInstance",
-        ];
```
rasmusgo, line 1268: Keeping track of the commands listed in the feature tag for `XR_VERSION_1_0` allowed me to get rid of this! 😄

rasmusgo:
This runs with api version 1.0 now. I'm not sure what, if anything, is missing from the 1.1 spec. The only thing that disappeared when when restricting the core commands to 1.0 was `xrLocateSpaces` from the high level bindings.

generator/src/main.rs
Ralith: Great catch!

generator/src/main.rs
```diff
@@ -192,27 +192,112 @@ impl Parser {
...
+                        extensions: vec![ext],
+                    },
+                );
+            } else {
+                eprintln!("ignoring extension with unlisted tag: {}", ext_name);
+            }
+        }
+    }
+
+    fn parse_required(
```
Ralith, line 267: Can you squash this into the commit that introduces feature parsing? Or better yet, have an initial commit that factors out the shared code, followed by one that introduces feature parsing? That would make review much easier.

generator/src/main.rs
```diff
@@ -1308,13 +1317,10 @@ impl Parser {
...
     #[allow(clippy::cognitive_complexity)] // TODO
     fn generate_hl(&self) -> TokenStream {
-        const BLACKLISTED_COMMANDS: [&str; 3] = [
-            "xrNegotiateLoaderRuntimeInterface",
-            "xrNegotiateLoaderApiLayerInterface",
-            "xrCreateApiLayerInstance",
-        ];
         let (instance_pfn_fields, instance_pfn_inits) = self.commands.iter().map(|(name, command)| {
-            if command.extension.is_some() || BLACKLISTED_COMMANDS.contains(&name.as_str()) {
+            if !command.feature.as_ref().map_or(false, |feature|
```
Ralith, line 1321: nit: push the `!` down for better readability

generator/src/main.rs
```diff
@@ -1308,13 +1317,10 @@ impl Parser {
...
+                feature.as_ref() == "XR_VERSION_1_0"
```
Ralith, line 1322: Explain in a comment why we don't want 1.1 stuff here too.

Ralith (changes requested):

generator/src/main.rs
rasmusgo: Done!

generator/src/main.rs
```diff
@@ -280,8 +313,7 @@ impl Parser {
                             } else if name.ends_with("EXTENSION_NAME") {
                                 let value = attr(&attributes, "value").unwrap();
                                 assert_eq!(&ext_name[..], &value[1..value.len() - 1]);
-                            } else {
-                                let value = attr(&attributes, "value").unwrap();
+                            } else if let Some(value) = attr(&attributes, "value") {
```
Ralith: Does removing the `unwrap` that was previously here put us at risk of missing future additions of new cases that the generator doesn't support?

generator/src/main.rs
```diff
@@ -259,6 +292,12 @@ impl Parser {
                         if let Some(extends) = attr(&attributes, "extends") {
                             const EXT_BASE: i32 = 1_000_000_000;
                             const EXT_BLOCK_SIZE: i32 = 1000;
+                            let ext_number = ext_number.unwrap_or_else(|| {
+                                attr(&attributes, "extnumber")
```
Ralith, line 304: Are we relying on the caller to pass down a number for extensions, and parsing it out here instead for features? Would it be simpler/more consistent to either handle both cases (`number` vs. `extnumber`) automatically here, or alternatively to parse the number out uniformly in the caller, making the argument non-optional?

Ralith (approved):

generator/src/main.rs
rasmusgo: Removing the unwrap should have come in a later commit but I missed to move it while rebasing.

generator/src/main.rs
Ralith: Merging as-is for the sake of downstream.

PR #159, opened 2024-05-07
Remove null terminators from other extensions

rasmusgo:
The field `other` in `ExtensionSet` currently handles null terminated strings in a weird way. It adds null terminators in `fn names` but doesn't remove them in `fn from_properties`. I got bit by this when checking for extensions by name and not having null terminators in my strings. I think it would make the most sense to not have null terminators in `String`s.

### DIFF
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -1518,6 +1518,7 @@ impl Parser {
                             #(#ext_set_inits)*
                             bytes => {
                                 if let Ok(name) = std::str::from_utf8(bytes) {
+                                    // This is probably always true but we check just to be sure.
                                     if bytes.last() == Some(&0) {
                                         out.other.push(name[0..(bytes.len() - 1)].into());
                                     } else {
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -1520,7 +1520,10 @@ impl Parser {
                             bytes => {
                                 let cstr = CStr::from_bytes_with_nul(bytes)
                                     .expect("extension names should be null terminated strings");
-                                let string = String::from_utf8_lossy(cstr.to_bytes()).to_string();
+                                let string = cstr
+                                    .to_str()
+                                    .expect("extension names should be valid UTF-8")
+                                    .to_string();
                                 out.other.push(string);
                             }
                         }
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -1495,6 +1495,7 @@ impl Parser {
 
             #![allow(clippy::wrong_self_convention, clippy::transmute_ptr_to_ptr)]
             use std::borrow::Cow;
+            use std::ffi::CStr;
             use std::mem::MaybeUninit;
             pub use sys::{#(#reexports),*};
             pub use sys::platform::{EGLenum, VkFilter, VkSamplerMipmapMode, VkSamplerAddressMode, VkComponentSwizzle};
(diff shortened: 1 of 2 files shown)


generator/src/main.rs
```diff
@@ -1518,7 +1518,11 @@ impl Parser {
                             #(#ext_set_inits)*
                             bytes => {
                                 if let Ok(name) = std::str::from_utf8(bytes) {
-                                    out.other.push(name.into());
+                                    if bytes.last() == Some(&0) {
```
Ralith, line 1521: Isn't this guaranteed to always be true?

Ralith (changes requested):
Good catch!

> I think it would make the most sense to not have null terminators in Strings.

That makes sense to me. We'll need to insert nulls in `fn names`, then.

generator/src/main.rs
rasmusgo: It might be. I just didn't want to depend on it. Do you think I should assume it's true and remove the test?

rasmusgo:
> Good catch!
> 
> > I think it would make the most sense to not have null terminators in Strings.
> 
> That makes sense to me. We'll need to insert nulls in `fn names`, then.

It is already doing that.

generator/src/main.rs
Ralith: I don't mind defensive programming, but it can be very confusing when an interface is used in a way inconsistent with its specified behavior. At least include an explanatory comment?

Ralith:
Revisiting this, there's a problem with the proposed approach: the most natural way to check for a name is to compare with `xr::sys::WHATEVER_EXTENSION_NAME`, which *does* include the null.

rasmusgo:
The `other` field is used for extensions not available as booleans and those don't have constants either. The names of those are copied from documentation or by listing available extensions. It's also not really a problem to have redundant null terminators in the extension names when listing required extensions. What bit me was that I was trying to have optional extensions by comparing my strings without null terminators with available extensions with null terminators.

rasmusgo:
The constants are also `[u8]` so I think it is fair that they contain `\0` but the variables of type `String` doesn't.

Ralith:
Fair enough.

generator/src/main.rs
```diff
@@ -1518,6 +1518,7 @@ impl Parser {
                             #(#ext_set_inits)*
                             bytes => {
                                 if let Ok(name) = std::str::from_utf8(bytes) {
+                                    // This is probably always true but we check just to be sure.
```
Ralith, line 1521: It's not "probably" true, it's guaranteed by the spec. Any implementation failing to do this is buggy.

generator/src/main.rs
```diff
@@ -1517,9 +1518,10 @@ impl Parser {
                         match crate::fixed_str_bytes(&ext.extension_name) {
                             #(#ext_set_inits)*
                             bytes => {
-                                if let Ok(name) = std::str::from_utf8(bytes) {
-                                    out.other.push(name.into());
-                                }
+                                let cstr = CStr::from_bytes_with_nul(bytes)
+                                    .expect("extension names should be null terminated strings");
+                                let string = String::from_utf8_lossy(cstr.to_bytes()).to_string();
```
Ralith, line 1523: I'm not sure it's a good idea to ever silently transform an extension name, since needing to do so again indicates a serious bug, and would break any comparisons. `CStr` also has some nice convenience methods for string conversion.
```suggestion
                                let string = cstr.to_str().expect("extension names should be valid UTF-8").to_string();
```

openxr/src/generated.rs
```diff
@@ -686,9 +687,10 @@ impl ExtensionSet {
                     out.htcx_vive_tracker_interaction = true;
                 }
                 bytes => {
-                    if let Ok(name) = std::str::from_utf8(bytes) {
-                        out.other.push(name.into());
-                    }
+                    let cstr = CStr::from_bytes_with_nul(bytes)
+                        .expect("extension names should be null terminated strings");
+                    let string = String::from_utf8_lossy(cstr.to_bytes()).to_string();
```
Ralith, line 692: As above

Ralith (changes requested):

Ralith (approved):
Thanks!

PR #160, opened 2024-05-17
Update ash to 0.38 in example

mezum:
Related #155
Ash removed all Builder structures and provided `Default` implementations to solve lifetime issues.
Example now uses them.

See also: https://github.com/ash-rs/ash/blob/master/Changelog.md#replaced-builders-with-lifetimessetters-directly-on-vulkan-structs


Ralith (approved):
Thanks!

PR #161, opened 2024-06-05
implement FB_render_model

awtterpip:
Implements the [`FB_render_model`](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_render_model) extension.

### DIFF
--- a/openxr/src/session.rs
+++ b/openxr/src/session.rs
@@ -359,6 +359,12 @@ impl<G> Session<G> {
     }
 
     pub fn load_render_model_fb(&self, model_key: RenderModelKeyFB) -> Result<Vec<u8>> {
+        let load_info = RenderModelLoadInfoFB {
+            ty: RenderModelLoadInfoFB::TYPE,
+            next: ptr::null_mut(),
+            model_key,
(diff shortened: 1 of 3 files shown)


openxr/src/session.rs
```diff
@@ -308,6 +310,102 @@ impl<G> Session<G> {
...
+                self.as_raw(),
+                &RenderModelLoadInfoFB {
+                    ty: RenderModelLoadInfoFB::TYPE,
+                    next: ptr::null_mut(),
+                    model_key,
+                },
+                &mut buffer,
+            ))?;
+            let mut out = Vec::with_capacity(buffer.buffer_count_output as usize);
+            loop {
```
Ralith, line 380: Could this use `get_arr`?

Ralith (changes requested):
Thanks!

openxr/src/session.rs
awtterpip: check it now, i switched to using get_arr

Ralith (approved):
Thanks!

Issue #42, opened 2020-08-16
Fails to link: unresolved externals from 'pathcch.h' (Windows)


GitHubJasper:
Tried to build the example and another project but always get this error.
`pathcch.h` is part of the win32 library but I should have all the necessary build tools for c++ installed (at least I think so the SDL crate works fine for example)
This is the error:
```
libopenxr_sys-9008a7c2938b17d8.rlib(filesystem_utils.obj) : error LNK2019: unresolved external symbol __imp_PathCchCanonicalize referenced in function "bool __cdecl FileSysUtilsGetCanonicalPath(class std::basic_string<char,struct std::char_traits<char>,class std::allocaferenced in function "bool __cdecl FileSysUtilsGetCanonicalPath(class std::basic_string<chartor<char> > const &,class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > &)" (?FileSysUtilsGetCanonicalPalocator<char> > &)" (?FileSysUtilsGetCanonicalPath@@YA_NAEBV?$basic_string@DU?$char_traits@Dth@@YA_NAEBV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@AEAV12@@Z)
```

Is there something I am missing?

Ralith:
This is likely a bug on our end; thanks for the report! OpenXR 1.0.11 added a use of `PathCchCanonicalize` and I need to work out what incantation is necessary to make rust link the definition.

Ralith:
Would you mind testing https://github.com/Ralith/openxrs/pull/43?

GitHubJasper:
Thanks for the quick reply!
But sadly it doesn't seem to work for me, get the same error :(
Can you reproduce it? Might be my setup...

GitHubJasper:
Just checked some older versions and `0.11.0` seems to work, might be because it uses an older OpenXR version
Thought that might be useful to know

VZout:
> Thanks for the quick reply!
> But sadly it doesn't seem to work for me, get the same error :(
> Can you reproduce it? Might be my setup...

I'm also running into this issue.

Ralith:
Thanks for the report. The call was added in OpenXR 1.0.11, so older versions of openxrs, which use older versions of the OpenXR loader, should work fine; I recommend rolling back to those until I have time to dig into this further, which should be within the week, unless someone else is able to investigate and make a PR.

Ralith:
[...]

PR #162, opened 2024-07-16
fix cross-compiling static loader

Timbals:
The build script used the `cfg!` macro to check the target os, but the "target" in a build script is actually the host. Use the `CARGO_CFG_TARGET_OS` environment variable instead.

Issue #42 [has been worked around](https://github.com/KhronosGroup/OpenXR-SDK-Source/pull/198), and then [fixed upstream](https://github.com/KhronosGroup/OpenXR-SDK-Source/pull/239), so the additional flag on windows is no longer necessary.

Linking the filesystem library is also done correctly upstream now: https://github.com/KhronosGroup/OpenXR-SDK-Source/blob/main/src/cmake/StdFilesystemFlags.cmake

---

I'm not sure the remaining flags are necessary anymore, but I don't have a macOS/FreeBDS/Linux setup to test it.

If anyone else wants to cross-compile the loader from Windows to Android, you probably have to use [ninja](https://ninja-build.org/) with the `CMAKE_GENERATOR=Ninja` environment variable because MSBuild adds incompatible compiler flags (`-fno-exceptions`, `-std=gnu++17`).

### DIFF
--- a/sys/build.rs
+++ b/sys/build.rs
@@ -16,13 +16,13 @@ fn main() {
         );
         println!("cargo:rustc-link-lib=static=openxr_loader");
 
-        if cfg!(any(target_os = "macos", target_os = "freebsd")) {
+        let target_os = std::env::var_os("CARGO_CFG_TARGET_OS")
+            .expect("missing CARGO_CFG_TARGET_OS environment variable");
+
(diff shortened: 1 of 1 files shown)


Ralith (approved):
Thanks!

PR #167, opened 2024-08-02
add headless graphics api

galister:
Currently, creating a headless session is needlessly complicated, and developers effectively need to lie that they use a graphics API that they are not actually intending to use.

Since no graphics API is a valid choice of graphics API, I made a implementation that simplifies creating such sessions.

The inner `create_session` will fail with the correct `XrResult` if headless session is attempted without first enabling `XR_MND_headless`.

Usage example:
```rust
let (session, _, _) = unsafe {
     instance.create_session(system, &xr::headless::SessionCreateInfo {})?
};
```

### DIFF
--- a/openxr/src/graphics/headless.rs
+++ b/openxr/src/graphics/headless.rs
@@ -1,0 +1,68 @@
+use crate::*;
+
+/// Used to create a session without graphics.
+///
+/// See [`XR_MND_headless`] for details.
+///
+/// [`XR_MND_headless`]: https://www.khronos.org/registry/OpenXR/specs/1.1/html/xrspec.html#XR_MND_headless
+pub enum Headless {}
+
+impl Graphics for Headless {
+    type Requirements = Requirements;
+    type SessionCreateInfo = SessionCreateInfo;
+    type SwapchainImage = HeadlessSwapchainImage;
+    type Format = HeadlessFormat;
+
+    fn raise_format(_: i64) -> Self::Format {
+        // used by enumerate_swapchain_formats, which returns empty
+        unreachable!()
+    }
+    fn lower_format(f: Self::Format) -> i64 {
+        // used by create_swapchain, which is not available in headless
+        match f {}
+    }
+
+    fn requirements(_instance: &Instance, _system: SystemId) -> Result<Requirements> {
+        Ok(Requirements {})
+    }
+
+    unsafe fn create_session(
+        instance: &Instance,
+        system: SystemId,
+        _info: &SessionCreateInfo,
+    ) -> Result<sys::Session> {
+        let info = sys::SessionCreateInfo {
+            ty: sys::SessionCreateInfo::TYPE,
+            next: std::ptr::null(),
+            create_flags: Default::default(),
+            system_id: system,
+        };
+        let mut out = sys::Session::NULL;
+        cvt((instance.fp().create_session)(
+            instance.as_raw(),
+            &info,
+            &mut out,
+        ))?;
+        Ok(out)
+    }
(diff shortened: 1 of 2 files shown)


openxr/src/graphics/headless.rs
```diff
@@ -0,0 +1,60 @@
...
+/// See [`XR_MND_headless`] for safety details.
+///
+/// [`XR_MND_headless`]: https://www.khronos.org/registry/OpenXR/specs/1.1/html/xrspec.html#XR_MND_headless
+pub enum Headless {}
+
+impl Graphics for Headless {
+    type Requirements = Requirements;
+    type SessionCreateInfo = SessionCreateInfo;
+    type SwapchainImage = u32;
+    type Format = i64;
```
Ralith, line 14: Are these types ever used in a headless session? Might be cleanest to define them as empty enums, if not (`enum HeadlessSwapchainImage {}`, `enum HeadlessFormat {}`).

Ralith (commented):
Thanks, this seems reasonable.

openxr/src/graphics/headless.rs
```diff
@@ -0,0 +1,60 @@
+use crate::*;
+
+/// Used to create a session without graphics.
+///
+/// See [`XR_MND_headless`] for safety details.
```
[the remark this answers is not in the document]
galister: there are some important points which are not safety related, so i removed the word `safety`

openxr/src/graphics/headless.rs
galister: did this and added comments as reminder

openxr/src/graphics/headless.rs
```diff
@@ -0,0 +1,68 @@
...
+    type SwapchainImage = HeadlessSwapchainImage;
+    type Format = HeadlessFormat;
+
+    fn raise_format(_: i64) -> Self::Format {
+        // used by enumerate_swapchain_formats, which returns empty
+        unreachable!()
+    }
+    fn lower_format(_: Self::Format) -> i64 {
+        // used by create_swapchain, which is not available in headless
+        unreachable!()
```
Ralith, line 22: nit: this could be a zero-branch `match` on the argument

openxr/src/graphics/headless.rs
```diff
@@ -0,0 +1,68 @@
...
+
+impl Graphics for Headless {
+    type Requirements = Requirements;
+    type SessionCreateInfo = SessionCreateInfo;
+    type SwapchainImage = HeadlessSwapchainImage;
+    type Format = HeadlessFormat;
+
+    fn raise_format(_: i64) -> Self::Format {
+        // used by enumerate_swapchain_formats, which returns empty
+        unreachable!()
```
Ralith, line 18: nit: This is technically reachable by user code, but probably indicates a bug. Maybe `panic!` with a helpful message?

Ralith (approved):

openxr/src/graphics/headless.rs
galister: one would need both of these to be true in order to hit this:
- the openxr runtime is not spec compliant and returns a non-empty set of formats from `xrEnumerateSwapchainFormats`
- the developer calls `enumerate_swapchain_formats`, disregarding headless session limitations

if the runtime is to spec, `xrEnumerateSwapchainFormats` will return an empty set and the inside of the `.map` that calls `raise_format` will never get executed.

Ralith: It's a public method, someone can just call it directly. Not really important since it's hidden, though.

PR #168, opened 2024-08-25
Change Action<Posef>::create_space to take a &Session

### LINKED ISSUE #130: Action<Posef>::create_space takes a Session instead of a &Session
It seems a little counterintuitive for this to consume a session vs just taking a reference to one, like the other Action associated functions.

Supreeeme:
Closes #130


Ralith (changes requested):
Thanks!

Ralith (approved):

PR #170, opened 2024-09-15
Fix UB in `Instance::supports_<prop>()`

### LINKED ISSUE #169: `Instance::supports_hand_tracking` might have undefined behavior
It seems that calling `get_system_properties` with `SystemProperties::next` pointing to a structure not supported by the reported extensions would not always result in an error. In fact it seems this is not documented in the OpenXR spec. In particular for the case of `Instance::supports_hand_tracking` we use `SystemHandTrackingPropertiesEXT::out()` and then call `assume_init()` after the `get_system_properties` succedes, but the field `supports_hand_tracking` might still contain random values.

I came across this while debugging a crash when using face tracking on a Quest Pro. Our code lead to test also for HTC lip  tracking support which somehow resulted as positive, in error. We were also using the `out()` method.

zarik5:
Closes #169

~~This solution works but it would be even more concise if the structs `sys::System<Prop>` would implement `Default`. The `Default` impl should already set the correct `ty`, and maybe the `out()` methods should base off the default value and return `T` instead of `MaybeUninit<T>`.~~
As per review, I switched to checking if the extension is present instead.

### DIFF
--- a/openxr/src/instance.rs
+++ b/openxr/src/instance.rs
@@ -151,6 +151,9 @@ impl Instance {
 
     #[inline]
     pub fn supports_render_model_loading(&self, system: SystemId) -> Result<bool> {
+        if self.exts().fb_render_model.is_none() {
+            return Err(sys::Result::ERROR_EXTENSION_NOT_PRESENT);
+        }
         unsafe {
             let mut render_model = sys::SystemRenderModelPropertiesFB::out(ptr::null_mut());
             let mut p = sys::SystemProperties::out(&mut render_model as *mut _ as _);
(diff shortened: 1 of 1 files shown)


Ralith:
Does the spec guarantee defined behavior when passing an extension struct from a disabled extension at all? I think it would be safer to check if the extension is enabled and bail out without making any XR calls at all if not.

zarik5:
Ok. Should I revert back to using T::out()?

Ralith:
That would be more consistent, yeah.

Ralith:
It looks like the answer to my question above [is actually "yes"](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#fundamentals-valid-usage-for-structure-pointer-chains):
> A runtime must ignore all unrecognized structures in a next chain, including those associated with an extension that has not been enabled.

So either approach seems fine. That said, the current form is concise and might make user error a bit more obvious, so I'm happy with it.

Ralith (approved):

zarik5:
I'm sorry I made a mistake. I opened another PR to check the correct extension

PR #172, opened 2024-09-20
implement XR_META_environment_depth

djefrey:
Implements the [XR_META_environment_depth](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_META_environment_depth) extension.


openxr/src/environment_depth.rs
```diff
@@ -0,0 +1,339 @@
...
+
+#[derive(Debug, Copy, Clone)]
+pub struct SystemEnvironmentDepthProperties {
+    pub supports_environment_depth: bool,
+    pub supports_hand_removal: bool,
+}
+
+impl Instance {
+    #[inline]
+    pub fn supports_environment_depth(&self, system: SystemId) -> Result<SystemEnvironmentDepthProperties> {
```
Ralith, line 23: This doesn't return a bool, so "supports" might be an odd name choice. Maybe `environment_depth_properties`?

Ralith (changes requested):

openxr/src/environment_depth.rs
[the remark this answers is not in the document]
djefrey: Done !

djefrey: Yes, that makes more sense.

pdiaz:
It would be great to have this extension. How can I help?

openxr/src/graphics/headless.rs
```diff
@@ -53,6 +53,12 @@ impl Graphics for Headless {
         // enumerate 0 formats, and so it's not possible to create a swapchain
         unreachable!();
     }
+
+    fn enumerate_depth_environment_swapchain_images(
+        _swapchain: &EnvironmentDepthSwapchain<Self>,
+    ) -> Result<Vec<Self::SwapchainImage>> {
+        unreachable!();
```
Ralith, line 60: Why is this unreachable? Document, or replace with an appropriate error.

openxr/src/graphics/d3d.rs
```diff
@@ -85,6 +85,27 @@ impl Graphics for D3D11 {
...
+        swapchain: &EnvironmentDepthSwapchain<Self>,
+    ) -> Result<Vec<Self::SwapchainImage>> {
+        let images = get_arr_init(
+            sys::SwapchainImageD3D11KHR {
+                ty: sys::SwapchainImageD3D11KHR::TYPE,
+                next: ptr::null_mut(),
+                texture: ptr::null_mut(),
+            },
+            |capacity, count, buf| unsafe {
+                (swapchain.fp().enumerate_environment_depth_swapchain_images)(
```
Ralith, line 99: What happens if this is called when the extension isn't present, or is present but not enabled?

Ralith (changes requested):

pdiaz:
@djefrey @Ralith Shall I push another merge request to help on this?

openxr/src/graphics/headless.rs
pdiaz: As it happens in enumerate_swapchain_images no swapchains are available in headless mode

Ralith:
Makes sense to me.

openxr/src/graphics/d3d.rs
djefrey: You cannot call this method without having the extension enabled and already used.

To use this method, you need an `EnvironmentDepthSwapchain`, which requires an `EnvironmentDepthProvider`.
If the extension is not present or loaded, `Session::create_depth_environment_provider` will panic with the message `'XR_META_environment_depth' not loaded.`

openxr/src/graphics/headless.rs
djefrey: I tried to create an headless session on a Quest 3, but it does not supports XR_MND_headless.
Futhermore, you need to create an EnvironmentDepthSwapchain to call this method, and this will probably fail in an headless session.

I'm considering adding this comment, what do you think ?
```
// as of today (March 2025), no runtime supports both XR_MND_headless and XR_META_environment_depth 
// if such a runtime exists in the future, you won't (probably) be able to create an EnvironmentDepthSwapchain
```

Ralith: It's not clear to me why we should assume these (or a future more general headless extension that might share this code) are mutually exclusive. Consuming information about the environment is independent of displaying graphics. Why not just handle the possibility?

Ralith: To be clear, I suspect the spec actually guarantees that these are mutually exclusive (since there's no way to communicate an image without one), but we should ensure that the failure is graceful, and document the reasoning with reference to the spec.

Issue #148, opened 2023-12-18
"requires target 'jsoncpp_interface' that is not in any export set"


valentinegb:
I have added `openxr` as a dependency to my project like so:

```toml
[dependencies]
openxr = { version = "0.17.1", features = ["static"] }
```

`cargo build` results in this error:

```
error: failed to run custom build command for `openxr-sys v0.9.3`

Caused by:
  process didn't exit successfully: `/Users/valentinebriese/Developer/bevy-openxr/target/debug/build/openxr-sys-2b99eab513ccc729/build-script-build` (exit status: 101)
  --- stdout
  CMAKE_TOOLCHAIN_FILE_aarch64-apple-darwin = None
  CMAKE_TOOLCHAIN_FILE_aarch64_apple_darwin = None
  HOST_CMAKE_TOOLCHAIN_FILE = None
  CMAKE_TOOLCHAIN_FILE = None
  CMAKE_GENERATOR_aarch64-apple-darwin = None
  CMAKE_GENERATOR_aarch64_apple_darwin = None
  HOST_CMAKE_GENERATOR = None
  CMAKE_GENERATOR = None
  CMAKE_PREFIX_PATH_aarch64-apple-darwin = None
  CMAKE_PREFIX_PATH_aarch64_apple_darwin = None
  HOST_CMAKE_PREFIX_PATH = None
  CMAKE_PREFIX_PATH = None
  CMAKE_aarch64-apple-darwin = None
  CMAKE_aarch64_apple_darwin = None
  HOST_CMAKE = None
  CMAKE = None
[...]

PR #173, opened 2024-09-21
Fix jsoncpp compilation error

RipleyTom:
Fixes https://github.com/Ralith/openxrs/issues/148.

It seems if the SDK detects jsoncpp on the system you can't build the loader as a static library unless you explicitly tell it to use the bundled jsoncpp(I think it's expecting to static link it which you can't do with the dynamic library it finds on the system).

### DIFF
--- a/sys/build.rs
+++ b/sys/build.rs
@@ -7,6 +7,7 @@ fn main() {
             .define("OpenGL_GL_PREFERENCE", "GLVND")
             .define("DYNAMIC_LOADER", "OFF")
             .define("CMAKE_INSTALL_LIBDIR", "lib")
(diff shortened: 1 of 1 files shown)


Ralith (commented):
Thanks! Maybe leave a comment in the file explaining why this is needed? Might also be worth opening an issue upstream.

RipleyTom:
While looking at their issues to see if someone had already reported it I stumbled upon:
https://github.com/KhronosGroup/OpenXR-SDK-Source/issues/481

I wish I'd looked before spending a hour figuring it out, put the issue url as a comment.

Ralith (approved):

PR #186, opened 2025-03-12
Upgrade OpenXR-SDK 1.1.40 and small fixes on generator

pdiaz:
This works towards upgrading OpenXR-SDK 1.1.40, not yet the last version but tries to lay down the path after making some decisions related to use a more consisting naming on types without removing the vendor suffix.


Ralith (approved):
Thanks!

PR #187, opened 2025-03-20
Don't acquire `xrEnumerateApiLayerProperties` when running as an API layer

yshui:
Honestly, this is not very nice, but the fact is both monado and SteamVR fails if their `xrGetInstanceProcAddr` functions are used to query "xrEnumerateApiLayerProperties". Monado returns `ERROR_NOT_SUPPORTED` if instance is not NULL, but will return a valid function if it is. On the other hand SteamVR returns `ERROR_HANDLE_INVALID` if instance _is_ NULL.

Given it doesn't really make sense to query API layers from an API layer, avoid all these troubles by not querying
"xrEnumerateApiLayerProperties" at all.

(Related: https://github.com/technobaboo/quark, this is a nicer version of https://github.com/Ralith/openxrs/commit/14824bfe883fde5d5e2c0e98b904a30d86d401f3)

### DIFF
--- a/generator/src/main.rs
+++ b/generator/src/main.rs
@@ -1330,8 +1330,18 @@ impl Parser {
                 pub #field_ident: pfn::#pfn_ident,
             };
             let c_name = c_name(name);
-            let init = quote! {
-                #field_ident: mem::transmute(entry.get_instance_proc_addr(instance, CStr::from_bytes_with_nul_unchecked(#c_name))?),
+
+            let init = if name != "xrEnumerateApiLayerProperties" {
+                quote! {
+                    #field_ident: mem::transmute(entry.get_instance_proc_addr(instance, CStr::from_bytes_with_nul_unchecked(#c_name))?),
+                }
+            } else {
+                quote! {
+                    #field_ident: entry
+                        .get_instance_proc_addr(instance, CStr::from_bytes_with_nul_unchecked(#c_name))
+                        .map(|s| unsafe { mem::transmute(s) })
+                        .unwrap_or(crate::stub_enumerate_api_layer_properties),
+                }
             };
             (field, init)
         }).unzip::<_, _, Vec<_>, Vec<_>>();
--- a/openxr/src/lib.rs
+++ b/openxr/src/lib.rs
@@ -59,6 +59,19 @@ pub const USER_HEAD: &str = "/user/head";
 pub const USER_GAMEPAD: &str = "/user/gamepad";
 pub const USER_TREADMILL: &str = "/user/treadmill";
 
+/// Stub for the `xrEnumerateApiLayerProperties` entry point when it's not provided by the runtime.
+///
+/// # Safety
+///
+/// No this is actually safe but we need the type to match.
+unsafe extern "system" fn stub_enumerate_api_layer_properties(
+    _a: u32,
+    _b: *mut u32,
+    _c: *mut sys::ApiLayerProperties,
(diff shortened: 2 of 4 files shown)


yshui:
That's fine too, but how do we communicate this failure to the caller? is the caller expected to check if `enumerate_api_layer_properties` is null before calling related functions?

Ralith:
As you say, it doesn't make sense for layers to call this, and for applications it should always be populated, so I don't think any special handling is required by callers. For robustness, on error we can replace the function with a stub that panics with a helpful message.

yshui:
Make sense. Does this apply to all functions that we fail to get, or just `xrEnumerateApiLayerProperties`?

sys/src/lib.rs
```diff
@@ -9,6 +9,17 @@ pub mod platform;
 #[cfg(feature = "mint")]
 mod mint_impls;
 
+/// # Safety
+///
+/// No this is actually safe but we need the type to match.
+pub unsafe extern "system" fn stub_enumerate_api_layer_properties(
```
Ralith, line 15: There's no reason for this to be public, or defined in a different crate than it's used in.

Ralith (changes requested):
> Does this apply to all functions that we fail to get, or just xrEnumerateApiLayerProperties?

It should only be applied to functions which may be absent in some conformant implementations under correct use.

sys/src/lib.rs
yshui: this is used in `openxr` though, do you want me to duplicate it in both places? or do you want me to add a `#[doc(hidden)]`

yshui: oh right, got a bit confused by where the generator code was going.

generator/src/main.rs
```diff
@@ -1330,8 +1330,18 @@ impl Parser {
...
             };
             let c_name = c_name(name);
-            let init = quote! {
-                #field_ident: mem::transmute(entry.get_instance_proc_addr(instance, CStr::from_bytes_with_nul_unchecked(#c_name))?),
+
+            let init = if name != "xrEnumerateApiLayerProperties" {
+                quote! {
+                    #field_ident: mem::transmute(entry.get_instance_proc_addr(instance, CStr::from_bytes_with_nul_unchecked(#c_name))?),
+                }
+            }else {
```
Ralith, line 1338: nit: `cargo fmt` please

openxr/src/lib.rs
```diff
@@ -59,6 +59,19 @@ pub const USER_HEAD: &str = "/user/head";
 pub const USER_GAMEPAD: &str = "/user/gamepad";
 pub const USER_TREADMILL: &str = "/user/treadmill";
 
+/// Stub for the `xrEnumerateApiLayerProperties` entry point when it's not provided by the runtime.
+///
+/// # Safety
+///
+/// No this is actually safe but we need the type to match.
+pub unsafe extern "system" fn stub_enumerate_api_layer_properties(
```
Ralith, line 67: This should not be `pub`.

Ralith (changes requested):

generator/src/main.rs
yshui: weirdly `cargo fmt` does not fix this.

Ralith: Huh. Sometimes it gives up if things get too complex, I guess we're hitting that.

Ralith (approved):

PR #192, opened 2025-04-03
Fix/silence lints

yshui:


Ralith (approved):
Thanks!

Ralith (changes requested):
Don't forget to actually run the generator.

Ralith:
Looks like the wrong lint name.

yshui:
and there are more cases. i should've run clippy before pushing

Ralith:
Thanks!

PR #196, opened 2025-05-25
Extension names as CStrs

Ralith:


Ralith:
TODO: Find a satisfying solution for matching extension names w/ conditional compilation. Maybe phf?

PR #217, opened 2026-04-18
add support for properly providing the Android `Activity` to the runtime

Timbals:
The context provided by `ndk-context` is not guaranteed to be an `Activity` context.

The `android-activity` crate has recently changed to providing an `Application` context to `ndk-context` instead (https://github.com/rust-mobile/android-activity/pull/229), but `XR_KHR_android_create_instance` requires a reference to the `Activity`.

The new `create_instance_android` function requires the user to pass in the `Activity` reference themselves.

Trying to use the normal `create_instance` function will produce the following error, which points users to the correct function:
```
error[E0599]: no method named `create_instance` found for reference `&openxr::Entry` in the current scope                                                                                                                                                              
  --> main.rs:73:26
   |
73 |     let instance = entry.create_instance(
   |                    ------^^^^^^^^^^^^^^^
   |
help: there is a method `create_instance_android` with a similar name
   |
73 |     let instance = entry.create_instance_android(
   |                                         ++++++++

For more information about this error, try `rustc --explain E0599`.                                                                                                                                                                                                    
error: could not compile `engine` (lib) due to 1 previous error                                                                                                                                                                                                        
Error: Command `cargo build --target aarch64-linux-android` had a non-zero exit code.
```

The `vulkan` example didn't compile for Android anymore, because the `ndk_glue::main` proc-macro doesn't parse c string literals. `ndk-glue` was deprecated some time ago (https://github.com/rust-mobile/ndk/issues/372), and should probably be replaced with `android-activity`.

### DIFF
--- a/openxr/src/entry.rs
+++ b/openxr/src/entry.rs
@@ -181,8 +181,11 @@ impl Entry {
         Ok(())
     }
 
-    /// Create an OpenXR instance with certain extensions enabled. Android support can be enabled by
-    /// setting `khr_android_create_instance` to `true`.
+    /// Create an OpenXR instance with certain extensions enabled.
+    ///
+    /// On Android, pass [`AndroidPlatformInfo`] as `platform_info`
+    /// and set `khr_android_create_instance` to `true`.
+    /// On other platforms, pass `()`.
     ///
     /// Most applications will want to enable at least one graphics API extension
     /// (e.g. `khr_vulkan_enable2`) so that a `Session` can be created for rendering.
@@ -191,6 +194,7 @@ impl Entry {
         app_info: &ApplicationInfo,
         required_extensions: &ExtensionSet,
         layers: &[&str],
+        platform_info: impl PlatformInfo,
     ) -> Result<Instance> {
         assert!(
             app_info.application_name.len() < sys::MAX_APPLICATION_NAME_SIZE,
@@ -216,29 +220,9 @@ impl Entry {
             .map(|x| x.as_ptr() as *const _)
             .collect::<Vec<_>>();
 
-        #[cfg(not(target_os = "android"))]
-        let next = ptr::null();
-        #[cfg(target_os = "android")]
-        let android_info = {
-            let context = ndk_context::android_context();
-
-            sys::InstanceCreateInfoAndroidKHR {
-                ty: sys::InstanceCreateInfoAndroidKHR::TYPE,
-                next: ptr::null(),
-                application_vm: context.vm(),
-                application_activity: context.context(),
-            }
-        };
-        #[cfg(target_os = "android")]
-        let next = if required_extensions.khr_android_create_instance {
-            &android_info as *const _ as _
-        } else {
-            ptr::null()
-        };
-
         let mut info = sys::InstanceCreateInfo {
             ty: sys::InstanceCreateInfo::TYPE,
-            next,
+            next: ptr::null_mut(),
             create_flags: Default::default(),
             application_info: sys::ApplicationInfo {
                 application_name: [0; sys::MAX_APPLICATION_NAME_SIZE],
@@ -258,9 +242,7 @@ impl Entry {
         );
         place_cstr(&mut info.application_info.engine_name, app_info.engine_name);
         unsafe {
-            let mut handle = sys::Instance::NULL;
-            cvt((self.fp().create_instance)(&info, &mut handle))?;
-
+            let handle = platform_info.create_instance(self, info)?;
             let exts = InstanceExtensions::load(self, handle, required_extensions)?;
             Instance::from_raw(self.clone(), handle, exts)
         }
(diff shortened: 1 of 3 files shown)


Timbals:
clippy failure in CI is due to a new Rust version and not related to the changes of this PR

Ralith:
If the activity context is still global state, why bother plumbing it through rather than fetching it internally the way we did before? I'm not a huge fan of the API special-case required.

Timbals:
There is no API to fetch the Activity internally. It's also not strictly global, as multiple activities can exist at the same time. The current activity could also get destroyed on configuration change, which would require re-creating the OpenXR instance with a new activity.
Admittedly an immersive OpenXR app doesn't really interact with the typical Android activity lifecycle in that way.

With the current implementation and `android-activity` 0.6.1 the session will never leave XR_SESSION_STATE_IDLE on Meta Quest devices.

Ralith:
You have updated the example to use `ndk_glue::native_activity().activity().cast()`, which fetches an activity from global state. It's not clear to me why (something equivalent to) this couldn't be substituted in directly for the existing purely internal activity lookup that you've identified as broken.

Timbals:
That's because the library doesn't depend on `ndk-glue`, only the examples do.
It also shouldn't depend on it, because that would force users of the library to use it for their glue code, when they should be using `android-activity`.

Ralith:
Does android-activity have an equivalent call? Could we use it internally?

Timbals:
It's [AndroidApp::activity_as_ptr](https://docs.rs/android-activity/0.6.1/android_activity/struct.AndroidApp.html#method.activity_as_ptr), but that requires the `AndroidApp` object, which is passed to the `android_main` entry point and not global.

Ralith:
Thanks for clarifying; that's a good reason to plumb it through (assuming they have a good reason for structuring it that way, anyway).

I'm still not a fan of having different platform-specific entry points that must be cfg-gated by the caller, though it's not immediately obvious to me what we should do instead. Maybe some sort of instance builder? That would be a much more robust pattern for arbitrary extension support in general, at the cost of being a much larger change...

Timbals:
I looked at how others deal with this:

- `hello_xr` example uses a [`IPlatformPlugin` interface](https://github.com/KhronosGroup/OpenXR-SDK-Source/blob/b8e48244207d9a623649e4de8cdbe288997194c1/src/tests/hello_xr/platformplugin.h)

- `winit` uses a builder, with [platform gated extension trait](https://docs.rs/winit/latest/winit/platform/android/trait.EventLoopBuilderExtAndroid.html#tymethod.with_android_app) and [panics when that is not called](https://github.com/rust-windowing/winit/blob/c4afadbfabf7b1e7989b40b493db1a4c7bd8ff4e/winit-android/src/event_loop.rs#L138-L141).

- `eframe` takes [the `AndroidApp` as part of an options struct](https://github.com/emilk/egui/blob/f342ab884797ee2a58b888bc0a35777a898f4717/crates/eframe/src/epi.rs#L407-L415) and [returns an error when that is not set](https://github.com/emilk/egui/blob/f342ab884797ee2a58b888bc0a35777a898f4717/crates/eframe/src/native/run.rs#L36-L42).

I think the reasonable options are:
1. two different functions for Android and non-Android (currently implemented)
2. builder pattern
    - could be used to implement the debug messenger for `XR_EXT_debug_utils` as well
    - should the current arguments be split into smaller methods on the builder? (`Builder::new(application_info)` or `Builder::new().application_info(application_info)` or `Builder::new().application_name(application_name).application_version(application_version)`)
3. introduce a `PlatformInfo` trait and always take a `impl PlatformInfo`
    - could be implemented for `()` on non-Android
    - `struct AndroidPlatformInfo` on Android, which is constructed with the activity JNI reference
4. always take a `Option<*mut c_void>`, and only require it to be `Some` on Android
    - this would require marking `create_instance` as `unsafe` for the Android case

I prefer option 3.

Timbals:
@Ralith I moved the passing of the `Activity` reference to the already existing `Entry::initialize_android_loader` function. That keeps `create_instance` the same for all platforms.

That means multiple/restarting activities are not supported, but I think that's fine for an XR application.

Ralith:
> introduce a PlatformInfo trait and always take a impl PlatformInfo

Good idea; that matches what we're doing for graphics APIs. For clarity and forwards-compat, we could use empty structs for platforms that don't (yet) need anything. I guess this could basically be seen as builders with platform-specific types.

> I moved the passing of the Activity reference to the already existing Entry::initialize_android_loader function

Hmm, I strongly prefer not to have hidden global state. Could this be a mutex-guarded field in `Entry`, at least? I think your other idea is ultimately the most attractive though.

In retrospect, I wonder why we have `initialize_android_loader` at all. From the doc it seems like we might as well call it on the user's behalf. Not an issue for this PR though.

Timbals:
> I think your other idea is ultimately the most attractive though.

Ok, i've implemented that now.
Notably, this could allow users to do arbitrary modifications to the `xrCreateInstance` call by implementing `PlatformInfo` themselves. Not sure if that's good, but I could seal the trait if we want to prevent that.

> In retrospect, I wonder why we have initialize_android_loader at all. From the doc it seems like we might as well call it on the user's behalf.

The original PR mentions this was to not change the error type for `Entry::load` and `Entry::load_from`.

openxr/src/entry.rs
```diff
@@ -258,9 +242,7 @@ impl Entry {
         );
         place_cstr(&mut info.application_info.engine_name, app_info.engine_name);
         unsafe {
-            let mut handle = sys::Instance::NULL;
-            cvt((self.fp().create_instance)(&info, &mut handle))?;
-
+            let handle = platform_info.create_instance(self, info)?;
             let exts = InstanceExtensions::load(self, handle, required_extensions)?;
```
Ralith, line 246: This has the safety invariant that `handle` must be a valid instance handle, so `PlatformInfo` must be an unsafe trait, documenting that invariant.

openxr/src/entry.rs
```diff
@@ -414,3 +396,78 @@ pub struct ApiLayerProperties {
...
+    ///
+    /// `info` must be valid to be passed to `xrCreateInstance`
+    unsafe fn create_instance(
+        &self,
+        entry: &Entry,
+        info: sys::InstanceCreateInfo,
+    ) -> Result<sys::Instance>;
+}
+
+#[cfg(not(target_os = "android"))]
```
Ralith: Does this actually need to be cfg-gated? Presumably the underlying runtime will fail in a graceful way if we omit the extension. I appreciate that it's nice for it to be a compile-time error, but it's also nice to not force downstream users to have cfg-gated code that only gets checked on specific targets and is hence very easy to break by accident.

openxr/src/entry.rs
```diff
@@ -414,3 +396,78 @@ pub struct ApiLayerProperties {
...
+#[cfg(target_os = "android")]
+impl AndroidPlatformInfo {
+    /// Creates a struct holding Android specific information for instance creation.
+    ///
+    /// Some information will also be fetched from [`ndk_context`],
+    /// so your NDK glue code has to initialize it.
+    ///
+    /// # Safety
+    ///
+    /// The `application_activity` pointer has to be a valid JNI reference to an `android.app.Activity`,
```
Ralith:
```suggestion
    /// The `activity` pointer has to be a valid JNI reference to an `android.app.Activity`,
```
There's nothing called `application_activity` here.

openxr/src/entry.rs
```diff
@@ -414,3 +396,78 @@ pub struct ApiLayerProperties {
...
+    /// as required by the [XR_KHR_android_create_instance] extension.
+    ///
+    /// If you use the `android-activity` crate, this will be the value provided by [`AndroidApp::activity_as_ptr`].
+    ///
+    /// [XR_KHR_android_create_instance]: https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_KHR_android_create_instance
+    /// [`AndroidApp::activity_as_ptr`]: https://docs.rs/android-activity/0.6.1/android_activity/struct.AndroidApp.html#method.activity_as_ptr
```
Ralith: This link to a specific version (and perhaps even the guidance towards a specific crate) is at high risk of becoming quite obsolete.

openxr/src/entry.rs
```diff
@@ -414,3 +396,78 @@ pub struct ApiLayerProperties {
...
+pub struct AndroidPlatformInfo {
+    activity: *mut std::ffi::c_void,
+}
+
+#[cfg(target_os = "android")]
+impl AndroidPlatformInfo {
+    /// Creates a struct holding Android specific information for instance creation.
+    ///
+    /// Some information will also be fetched from [`ndk_context`],
+    /// so your NDK glue code has to initialize it.
```
Ralith, line 437: Would it be possible/better to take that stuff as an argument as well?

Ralith (changes requested):

Ralith:
> Notably, this could allow users to do arbitrary modifications to the xrCreateInstance call by implementing PlatformInfo themselves. Not sure if that's good, but I could seal the trait if we want to prevent that.

I don't think there's anything wrong with that, so long as the safety invariants are clear. Preventing a sufficiently determined user from blowing their foot off is not a priority.

Ralith:
> The original PR mentions this was to not change the error type for Entry::load and Entry::load_from.

Perhaps we could do it in `create_instance`, though?

openxr/src/entry.rs
Timbals: done

openxr/src/entry.rs
Timbals: Ok, I personally would have preferred a compile-time error, but I've changed it.

openxr/src/entry.rs
Timbals: Thanks, I missed that while renaming

openxr/src/entry.rs
Timbals: Ok, I've removed it and will probably do a follow-up PR to update the examples to use `android-activity` since that is what everyone seems to use now instead of `ndk-glue`

Timbals:
>> The original PR mentions this was to not change the error type for Entry::load and Entry::load_from.

> Perhaps we could do it in create_instance, though?

No, it has to precede all other calls to OpenXR, including `Entry::enumerate_extensions` and `Entry::enumerate_layers`.
I think this should be called implicitly when creating the `Entry`. Changing the error types should be fine, it's just a breaking change.

openxr/src/entry.rs
Timbals: We could, but that is truly global state and we require it in `initialize_android_loader` as well, so the `ndk-context` initialization has to happen anyway.

Ralith (approved):

Ralith:
> I think this should be called implicitly when creating the Entry. Changing the error types should be fine, it's just a breaking change.

Yeah, reviewing the spec I think you're right. It was nice having `Entry::linked` be infallible; oh well.

PR #218, opened 2026-04-22
make loader initialization implicit

Timbals:
The `initialize_android_loader` function is called automatically when creating an `Entry` now. All functions to create an `Entry` now return an error when the initialization fails.

As discussed in https://github.com/Ralith/openxrs/pull/217#issuecomment-4298650980

### DIFF
--- a/openxr/src/entry.rs
+++ b/openxr/src/entry.rs
@@ -30,8 +30,8 @@ impl Entry {
     /// actually linked into the binary, e.g. by enabling the `static` feature or manually linking
     /// to an external loader or implementation.
     #[cfg(feature = "linked")]
-    pub fn linked() -> Self {
-        Self {
+    pub fn linked() -> Result<Self> {
+        let entry = Self {
             inner: Arc::new(Inner {
                 raw: RawEntry {
                     get_instance_proc_addr: sys::get_instance_proc_addr,
@@ -43,7 +43,10 @@ impl Entry {
                 #[cfg(feature = "loaded")]
                 _lib_guard: None,
             }),
-        }
+        };
+        #[cfg(target_os = "android")]
+        entry.initialize_android_loader()?;
+        Ok(entry)
     }
 
     /// Load entry points at run time from the dynamic library `openxr_loader` according to the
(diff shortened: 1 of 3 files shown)


openxr/src/entry.rs
```diff
@@ -344,26 +360,31 @@ pub struct RawEntry {
...
-pub struct LoadError(libloading::Error);
-
-#[cfg(feature = "loaded")]
-impl fmt::Debug for LoadError {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        self.0.fmt(f)
-    }
+#[derive(Debug)]
+pub enum EntryError {
+    Load(libloading::Error),
```
Ralith: This should wrap, rather than replace, `LoadError`. That way breaking changes in `libloading` are not breaking changes in this crate.

Ralith (approved):
Thanks!

openxr/src/entry.rs
Timbals: done

openxr/src/entry.rs
```diff
@@ -158,9 +172,9 @@ impl Entry {
         unsafe { get_instance_proc_addr_helper(self.fp().get_instance_proc_addr, instance, name) }
     }
 
-    /// Initialize Android loader. This must be called before `create_instance()`.
+    /// Initialize Android loader. This must be called before any other OpenXR call.
     #[cfg(target_os = "android")]
-    pub fn initialize_android_loader(&self) -> Result<()> {
+    fn initialize_android_loader(&self) -> Result<()> {
```
Ralith, line 177: [The spec](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_KHR_loader_init) says:
> If this extension is supported, its use may be required on some platforms and the use of the [xrInitializeLoaderKHR](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#xrInitializeLoaderKHR) function must precede other OpenXR calls except [xrGetInstanceProcAddr](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#xrGetInstanceProcAddr).

I think this means we should call `LoaderInitKHR::load` unconditionally, and then call `initialize_loader` if it succeeds and proceed normally otherwise, without any of these `target_os = "android"` gates.

Ralith (changes requested):

openxr/src/entry.rs
Timbals: I'm not sure how to do that. We can't construct a valid `LoaderInitInfoAndroidKHR` on non-Android platforms because we don't have the data from `ndk-context`.
`LoaderInitKHR::load` will always succeed in practice on all platforms, because the loader implements it, so that doesn't indicate if platform specific information is required.

Ralith: On closer examination it's all pretty clearly android specific, I guess this is fine.

Ralith (approved):

PR #219, opened 2026-06-03
Allow passing a proper activity pointer to the android loader

MaximilianHausen:
Extends the PlatformInfo trait introduced in #217 to also handle android loader initialization.

Previously, only `create_instance` used the user-provided activity pointer from `AndroidPlatformInfo`, while `initialize_android_loader` still relied on ndk_context. That should actually work based on the spec ([XR_KHR_loader_init_android](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_KHR_loader_init_android) just requires any `android.content.Context`), but it wasn't enough when testing android-activity 0.6.1 on my quest 3.

I decided against storing the `PlatformInfo` from the constructors in `Entry` because that would make `Entry` `!Sync` by default and I am not 100% sure if implementing it manually would be correct here. Keeping it like this is also more flexible because it allows users to pass different pointers to `init_loader` and `create_instance` (I don't see when that would be neccessary, but it could be useful for debugging).

### DIFF
--- a/openxr/examples/hello.rs
+++ b/openxr/examples/hello.rs
@@ -2,11 +2,16 @@ use openxr as xr;
 
 #[cfg_attr(target_os = "android", ndk_glue::main)]
 fn main() {
+    #[cfg(not(target_os = "android"))]
+    let platform_info = ();
+    #[cfg(target_os = "android")]
+    let platform_info =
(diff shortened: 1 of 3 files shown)


Ralith (approved):
Thanks! I think this makes sense. It'd be nice to have better ergonomics by saving the data, but if it were that easy I assume OpenXR would do so itself, so probably good to respect the underlying pattern here.

PR #220, opened 2026-06-25
Remove `ndk-context` dependency

Timbals:
#218 started taking in some Android specific information using a `PlatformInfo` for instance creation, and #219 expanded that to also use the `PlatformInfo` for initializing the loader on Android. `ndk-context` is still used to get the `JavaVM` pointer.
That means we currently have the downsides of both
- `ndk-context` global has to be initialized
- explicit platform specific arguments have to be passed

This PR removes the `ndk-context` dependency and passes all data explicitly using the `PlatformInfo`, including the `JavaVM` pointer.

I also removed the `Sized` bound for `PlatformInfo` so that user code can use `dyn PlatformInfo`. That's not relevant for the examples in this repo, but for my usage this avoids cfg-gating in multiple places.

### DIFF
--- a/openxr/Cargo.toml
+++ b/openxr/Cargo.toml
@@ -32,9 +32,6 @@ ctrlc = "3.1.5"
 [target.'cfg(target_os = "android")'.dev-dependencies]
 ndk-glue = "0.7"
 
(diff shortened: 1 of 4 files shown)


Ralith (approved):
Thanks!

PR #223, opened 2026-08-23
add missing Android libraries when using `static` feature

Timbals:
This fixes two things when using the `static` feature on Android:

- Links shared libraries required by the loader. Similar to #207. The libraries are `libandroid.so` and `liblog.so`.
  `libandroid.so` is normally linked by `ndk-sys`, but I had a recent project that didn't include that.
  `liblog.so` is linked by the `std`. It's included for completeness, in case `std` changes, or this library ever wants to build without `std`.
- Links the C++ runtime library that is used and needed by the loader. We previously linked to `libstdc++.so`, [which only includes `new` and `delete` on Android](https://developer.android.com/ndk/guides/cpp-support#system). In practice this meant the user had to manually link a C++ runtime library (equivalent to the new `OPENXRS_ANDROID_STL=none`). This wasn't documented anywhere either and probably only worked accidentally when another library also linked the C++ runtime library (e.g. `cpal` before version 0.16).
  By default the `static` feature will now also statically link the C++ runtime library.
  I've added a new environment variable `OPENXRS_ANDROID_STL` that can be used to change the linking behavior for the C++ runtime library when a statically linked C++ runtime library is not appropriate. It's described in the README.md.

I've had these changes in the `build.rs` of my own projects for a while now, but they are more appropriate to be included here.


Ralith (approved):
Thanks! I especially appreciate the thorough documentation and cites.

PR #225, opened 2026-09-15
add .1 suffix to OpenXR loader library path for linux systems

Mr-Zero88:
Should i add the suffix only if target is linux or is this also fine?


Ralith (approved):
Thanks!

Ralith:
> Should i add the suffix only if target is linux or is this also fine?

For comparison, ash does:

```rs
        #[cfg(windows)]
        const LIB_PATH: &str = "vulkan-1.dll";

        #[cfg(all(
            unix,
            not(any(
                target_os = "macos",
                target_os = "ios",
                target_os = "android",
                target_os = "fuchsia",
                target_env = "ohos"
            ))
        ))]
        const LIB_PATH: &str = "libvulkan.so.1";

        #[cfg(any(target_os = "android", target_os = "fuchsia", target_env = "ohos"))]
        const LIB_PATH: &str = "libvulkan.so";

        #[cfg(any(target_os = "macos", target_os = "ios"))]
        const LIB_PATH: &str = "libvulkan.dylib";
```

So as written, I guess this will break android (i.e. Quest), so we should at least handle that.

Ralith (changes requested):
See ^

Mr-Zero88:
i would just put the .s suffix to only apply to the linux target and keep the default like before. then it will not break any other platform.

Ralith (approved):
Might be wrong on BSD, but there probably aren't any OpenXR implementations there yet.

